PR #1, opened 2016-01-26
Initial port steps

zachmullen:
* Remove girder plugin code (will move into girder/girder)
* Rename romanesco -> girder_worker


jcfr:
:+1:

zachmullen:
Just talked with @mgrauer , I think we should go ahead and review and merge this. There's a lot of work to do on the documentation, but given that it should be much more than simply renaming "romanesco", I think it's best to do it on a subsequent topic branch. Since nobody's using this repo yet and our docs aren't deployed yet, I don't think that will be problematic.

jcfr:
:+1:

mgrauer:
LSaneTM.  I expect there will be more changes to come, but this passes tests and has good coverage on a dashboard.

PR #2, opened 2016-01-27
Scope the entire script so it executes in one step

jeffbaumes:
Also ensure the script does not require a trailing newline


zachmullen:
@jeffbaumes could you rebase this? The file in question has moved to its new namespace.

jeffbaumes:
Ok I think I did that right? Should be ready now.

zachmullen:
LGTM

PR #5, opened 2016-02-15
Pass an optional reference from the JobManager to uploaded files.

manthey:
This is dependent on an update to the girder client to v 1.1.2.


zachmullen:
LGTM

PR #6, opened 2016-02-17
Added new types to support translation from xml spec of slicer CLI

cdeepakroy:
Added new types -- integer, integer_list, number_list, string_list - to support translation from [xml spec of slicer execution model CLI](https://www.slicer.org/slicerWiki/index.php/Slicer3:Execution_Model_Documentation#XML_Schema)

### DIFF
--- a/girder_worker/format/integer/validate_json.json
+++ b/girder_worker/format/integer/validate_json.json
@@ -1,0 +1,7 @@
+{
+    "inputs": [{"name": "input", "type": "integer", "format": "json"}],
+    "outputs": [{"name": "output", "type": "boolean", "format": "boolean"}],
+    "script": "output = isinstance(input, (str, unicode))",
+    "extensions": ["integer-json"],
+    "mode": "python"
+}
--- a/tests/integer_list_test.py
+++ b/tests/integer_list_test.py
@@ -1,0 +1,81 @@
+import girder_worker
+import unittest
+
+
+class TestIntegerList(unittest.TestCase):
+
+    def setUp(self):
+        self.analysis = {
+            "name": "concatenate",
+            "inputs": [
+                {
+                    "name": "a",
+                    "type": "integer_list",
+                    "format": "integer_list",
+                    "default": {
+                        "format": "json",
+                        "data": "[0]"
+                    }
+                },
+                {
+                    "name": "b",
+                    "type": "integer_list",
+                    "format": "integer_list"
+                }
+            ],
+            "outputs": [{"name": "c",
+                         "type": "integer_list", "format": "integer_list"}],
+            "script": "c = a + b",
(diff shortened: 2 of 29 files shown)


girder_worker/format/integer/validate_json.json
```diff
@@ -0,0 +1,7 @@
+{
+    "inputs": [{"name": "input", "type": "integer", "format": "json"}],
+    "outputs": [{"name": "output", "type": "boolean", "format": "boolean"}],
+    "script": "output = isinstance(input, (str, unicode))",
+    "extensions": ["integer-json"],
```
zachmullen: What does extensions do?

zachmullen:
I'm fine with these new types, but I'd like @jeffbaumes to weigh in on whether this is in line with how he thinks the type/format system should look.

girder_worker/format/integer/validate_json.json
danlamanna: `extensions` are just a way for Flow to infer what type/format something is. If there are conflicts (i.e. tables and graphs could both theoretically be stored as a csv) then Flow forces the user to resolve the conflict.

tests/integer_list_test.py
```diff
@@ -0,0 +1,81 @@
...
+                    "format": "integer_list"
+                }
+            ],
+            "outputs": [{"name": "c",
+                         "type": "integer_list", "format": "integer_list"}],
+            "script": "c = a + b",
+            "mode": "python"
+        }
+
+    def test_integer(self):
```
jcfr: should this be named `test_interger_list` ?

jcfr:
:+1:

tests/integer_list_test.py
cdeepakroy: I just followed the naming convention used for the [rest of the tests](https://github.com/girder/girder_worker/tree/master/tests) in girder_worker.

jeffbaumes:
I think this follows everything appropriately. This goes even further down the path of a flat list of types without complex typing logic, hence the explicit code for `number_list` instead of having some girder_worker `listof` construct to generally make a list type via something like `listof(number)`. I prefer staying on this simpler path - otherwise we will end up needing to use (or worse, create) an advanced typing system like [datashape](http://datashape.readthedocs.org/en/latest/) which would be heavyweight and complex - which is against what we are trying to do here I think.

All that to say LGTM.

cdeepakroy:
Thanks for the input everyone! Can someone merge it? I don't seem to have rights.

jcfr:
Et voila. 

(Someone with admin access would have to grant push access.)

zachmullen:
@cdeepakroy I just invited you to the developers team, which should give you write access to all repos in this org.

PR #9, opened 2016-03-11
Use docker-gc to clean up outdated containers & images

zachmullen:


zachmullen:
I just verified that this works using my COVALIC deployment. An extraneous image with no containers was removed, and all containers that had been used in the last hour were preserved. PTAL

zachmullen:
@brianhelba this contains some new features that will be helpful for hosting the ISBI challenge. In general, this will now support using images that exist locally but not in docker hub.

girder_worker/plugins/docker/docker-gc
```diff
@@ -0,0 +1,267 @@
+#!/bin/bash
+
+# Taken from https://github.com/spotify/docker-gc
```
jcfr: Could you be more explicit by for example specifying the exact revision. 

E.g:

Taken from spotify/docker-gc@fa99444

docs/plugins.rst
```diff
@@ -58,6 +59,10 @@ The temporary directory for the Girder Worker task is mapped into the running co
 under the directory ``/data``, so any files that were fetched into that temp directory
 will be available inside the running container at that path.
 
+By default, the image you specify will be pulled using the ``docker pull`` command.
+In some cases, you may not want to perform a pull, and instead want to rely on the
+image already being present on the worker system. If so, set ``pull_image`` to false.
```
jcfr: Can we specify that disabling pulling is interesting only if only the image name of `image:tag` (e.g `foo:latest`) is specified. In that case, it will prevent systematic update to the latest version. That said, in the scientific context, explicitly specifying the digest of the image is recommended. It ensures the execution context is always the same. See https://docs.docker.com/engine/reference/run/#image-digest

zachmullen: I don't know enough about what you're talking about to add that to the docs, but if you add a sentence or two about that it would be cool :)

girder_worker/plugins/docker/docker-gc
zachmullen: Fixed

jcfr:
@zachmullen Et voila. I added a note in 0a1eb13 

Cc: @thewtex

zachmullen:
@jcfr your changes look good to me, are you OK with the rest of this branch?

cpatrick:
lgtm.

zachmullen:
Heads up, I added one extra commit to add a new feature that lets the task spec add additional command line options to `docker run`. This will allow clients to send tasks that can, for example, disable network traffic to and from the container.

PR #14, opened 2016-03-29
Setting flake8 width to 80 characters

jeffbaumes:
Fixes #10. Down from 90 characters. Flake8 config now exactly matches Girder.

PR #16, opened 2016-03-29
Single quotes

### LINKED ISSUE #15: Normalize to single quotes
There is not a flake8 rule for this but we can at least go through and convert all double quotes to single quotes where possible/reasonable. Currently the codebase looks a bit disjointed.

jeffbaumes:
Depends on #14. Fixes #15.


tests/spec_test.py
```diff
@@ -1,4 +1,4 @@
-"""Tests for core spec objects."""
+'''Tests for core spec objects.'''
```
msmolens: We explicitly [ignore](https://github.com/girder/girder_worker/blob/3df3b2edf6f0718b3c79a81770afe824abf7aea9/tests/flake8.cfg#L56) the flake8-docstrings `D300 - Use """triple double quotes"""` test, but is there a reason to change this file's docstrings to use single quotes?

jeffbaumes: No, good catch. It was a find/replace error. I'll search for these and push an update.

jeffbaumes:
@msmolens I've updated the triple quotes so that the command

```
git grep "'''"
```

produces nothing. I feel we should consistently use double-quotes for this docstrings and multiline strings, though as you say it is not enforced.

msmolens:
LGTM

PR #17, opened 2016-03-30
Enforce more of PEP257 -- Docstring Conventions

msmolens:
Enforce the following warnings from flake8-docstrings for [PEP257](https://www.python.org/dev/peps/pep-0257/):

    D208 - Docstring over-indented.
    D209 - Put multi-line docstring closing quotes on separate line.
    D300 - Use """triple double quotes""".

### DIFF
--- a/girder_worker/specs/workflow.py
+++ b/girder_worker/specs/workflow.py
@@ -80,14 +80,16 @@ class Workflow(MutableMapping):
         self.__graph__.add_edge(t1, t2, metadata)
 
     def connect_tasks(self, *args, **kwargs):
-        """Connect two tasks together specifying which input ports connect
+        """
+        Connect two tasks together specifying which input ports connect
         to which output ports. This function takes either 1, 2 or 3 arguments.
         In the case of one argument it expects a list of tuples where the tuples
         are the first, second and third arguments to connect_tasks. If there are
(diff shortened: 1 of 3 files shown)


zachmullen:
LGTM

jeffbaumes:
LGTM also. In general can we start to have a strategy for syncing Python code style on Kitware projects? @zachmullen perhaps you can think about this - might just be a matter of having a meta-task to update flake8 config files across repos once there is consensus that a certain code style is a good idea. Or perhaps we don't want to enforce normalization of Python style.

zachmullen:
I think git submodules would be the most obvious solution, but a lot of people dislike using them...

brianhelba:
+1 to removing duplication of code and enforcing Kitware styles.

IMO, Git submodules in general are a fine idea with a terrible implementation. If we start to require them in our build process, some people are going to feel pain.

Rather than girder-worker having it's own testing stack, would it make sense to make Girder a requirement for running the tests, then just test girder-work as if it were a Girder plugin (that is, initiated through Girder's testing stack). I don't even necessarily like this idea, but it's an option.

As a much better idea, could we refactor Girder's common test infrastructure (for being invoked via CTest, for handling both Python tests and JS tests, for running style checkers, etc.), into its own library (with an associated Pip package), that any Kitware project could then depend on?

PR #18, opened 2016-04-07
Add additional extensions for Newick trees

danlamanna:
@curtislisle @jeffbaumes 

This also removes a test that was relying on an external API that no longer seems to be functioning.


curtislisle:
LGTM

jeffbaumes:
LGTM2

PR #20, opened 2016-04-18
Output files from docker tasks

zachmullen:


girder_worker/plugins/docker/executor.py
```diff
@@ -118,6 +118,26 @@ def _docker_gc(tempdir):
...
 
+def validate_task_outputs(task_outputs):
+    """
+    This is called prior to fetching inputs to make sure the output specs are
+    valid. Outputs in docker mode can result in side effects, so it's best to
+    make sure the specs are valid prior to fetching.
+    """
+    for name, spec in task_outputs.iteritems():
+        if spec.get('target') == 'filepath':
+            if 'path' not in spec:
```
cdeepakroy: Why not assume the path to be `os.path.join('/data', spec['name'])`?

cdeepakroy: Are you asking them to specify the `path` in the task spec, so people could create sub-directories under `/data` to avoid file name conflicts?

cdeepakroy: Also if you do want them to specify `path`, maybe you should ask them to specify it in the binding spec? I feel this is kind of a binding (where will the data be produced) specification rather than the task specification.

cdeepakroy: Maybe just ask for the relative path and you can internally prepend/join it with `/data` .

zachmullen: > Are you asking them to specify the path in the task spec, so people, if needed, can create sub-directories under /data to avoid file name conflicts?

Yes, in fact I'd like them to be able to point to a directory and have the whole thing uploaded to girder recursively if they want.

> Maybe just ask for the relative path and you can internally prepend/join it with /data

I will make it so it works with either option; if an absolute path is specified, it must start with "/data/", but if a relative path is specified it will be assumed as relative to /data.

jeffbaumes:
Could we go ahead and update the docs on this branch as well? I see some doc-like stuff in the discussion, such as:

> I will make it so it works with either option; if an absolute path is specified, it must start with "/data/", but if a relative path is specified it will be assumed as relative to /data.

girder_worker/plugins/docker/executor.py
zachmullen: > Also if you do want them to specify path, maybe you should ask them to specify it in the binding spec? I feel this is kind of a binding (tell me where will the data be produced) specification rather than the task specification.

I think the task output is the more appropriate place for it. The task is in charge of declaring *what* it produces, the output binding declares *how* it should be handled afterward. I think declaring "I produce a file named foo" is declaring *what* you are producing, and it allows us to handoff nicely and decouple from the push logic.

zachmullen:
@cdeepakroy @jeffbaumes I think I addressed everything, PTAL. I think I will do #21 on a subsequent topic for cleanliness.

jeffbaumes:
Docs LGTM. Also helped me understand how this works. I'll let @cdeepakroy approve on the functionality side.

girder_worker/plugins/docker/executor.py
cdeepakroy: @zachmullen I understand now. 

I also like that you have made it such that if path is not provided then it defaults to '/data/ + name'.

cdeepakroy: Oh! i just tested it on my end. 

If `path` is not provided it raises TaskSpecValidationError('Docker filepath outputs must specify a "path" field.'.

Why not make it such that if `path` is not provided it internally assumes it to be `/data/ + spec[name]`?

zachmullen: I did make that change in commit 9b90af0, did you test on this commit or with the new ones as well?

cdeepakroy: My bad. Forgot to git pull.

Now that i am modifying my code to provide the path, i feel the binding spec is more appropriate.
I feel the taskSpec is supposed to say `this task will provide an output file of type blah`. And the bindingSpec is supposed to say `this output file should be stored at xyz folder in girder with name abc`

cdeepakroy: @jeffbaumes can we make `task['input'] and task['outputs']` a dictionary instead of a list. While generating the task spec automatically it will make it computationally faster to find and modify a specific input or output. Otherwise finding an input in the list is an `O(n)` operation.

jeffbaumes: @cdeepakroy this was discussed early on. The issue with a dictionary is that it is unordered. If you want to present inputs in a UI, they will be in an unordered jumble if the inputs are a dictionary. If you need to lookup inputs or outputs by name/id, you can always quickly make a map in `O(n)` time and then to quick lookups.

Also, I don't see input and output lists being long enough that `O(n)` lookups will cause any sort of overall performance issue.

So that's the rationale. If there was a good way to create convenience method that would help ease this I'd be all for it. There was some work on a Pythonic dict subclass for Girder Worker tasks that could have convenience methods like this - perhaps we could revisit that.

cdeepakroy: Ok!

@zachmullen Its good to merge by me now.

PR #22, opened 2016-04-19
New console script entrypoint for managing config file

zachmullen:
This allows users to much more easily manage their worker config
file, without having to dig down into install directories, copy
and paste the dist file to local file, etc.

This also makes it so plugins are loaded lazily rather than
immediately at top-level module import time, to avoid polluting
standard output with plugin load messages.

### DIFF
--- a/girder_worker/__init__.py
+++ b/girder_worker/__init__.py
@@ -20,7 +20,6 @@ config.read([os.path.join(PACKAGE_DIR, f) for f in _cfgs])
 
 # Maps task modes to their implementation
 _task_map = {}
-_plugins_loaded = False
 
 
 class TaskSpecValidationError(Exception):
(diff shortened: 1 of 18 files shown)


girder_worker/__init__.py
```diff
@@ -209,6 +219,8 @@ def run(task, inputs=None, outputs=None, auto_convert=True, validate=True,
         was provided. Instead, those outputs will be saved to that URI and
         the output binding will contain the location in the ``'uri'`` field.
     """
+    load_plugins()
+
```
jeffbaumes: I feel like `load_plugins()` should go anywhere that might need plugins loaded and not just the `run` function. Other possible entry points are `isvalid`, `load`, `convert`, `format.has_converter`, `format.get_validator_analysis`, and `format.converter_path`. See my related comment on tests.

girder_worker/plugins/docker/tests/docker_test.py
```diff
@@ -42,6 +42,8 @@ def _mockOsRead(fd, *args, **kwargs):
 
 
 def setUpModule():
+    girder_worker.load_plugins()
+
```
jeffbaumes: If `load_plugins` is lazily-loaded correctly (see prior comment), `load_plugins` should not be needed on any of the tests.

jeffbaumes:
@zachmullen see my inline comments. The CLI will be very useful but I have questions on the lazy plugin initialization.

zachmullen:
Your points about plugin loading are sound. I can't come up with a good way to avoid loading them at module load time, so I'll just go back to doing that, but make it silent by default since the point was to avoid polluting stdout.

zachmullen:
We may want to think about this, though, in case we have any use cases in the future where one might want to use code inside the `girder_worker` package without loading plugins.

jeffbaumes:
I agree that it seems cleaner to leave the plugin loading on import for now. If you can think of a task to split out for revisiting this it might be good to track. I cannot currently think of when someone would want to use `girder_worker` without its plugins. If you don't want any plugins you can always use your CLI to remove them all from your config.

zachmullen:
Removed the changes related to plugin loading.

zachmullen:
@jeffbaumes does this look Ok to you now that I reverted the plugin load behavior?

jeffbaumes:
All the commands seem to work for me except `rm`. It always appears to succeed without an error but does not seem to remove anything if I do a subsequent `list`.

Also needs to be added as a short section in the docs, which could be in this PR or a subsequent one.

jeffbaumes:
A future addition might be to have a special command for adding to or removing from the plugin list, instead of needing to set the full plugin list each time. But that is not needed in this PR and could potentially be overkill.

zachmullen:
> All the commands seem to work for me except rm. It always appears to succeed without an error but does not seem to remove anything if I do a subsequent list.

Which option are you trying this on? We only modify the local config, so if it's one of the defaults from the dist config, it will still appear.

jeffbaumes:
Ah I think I get it. So the defaults will shine through even if you've done `rm` - another reason for docs since someone (e.g. me) might think `rm` would make an option disappear. Other than docs (which can be on another PR) LGTM.

zachmullen:
@jeffbaumes I just added some docs

jeffbaumes:
Docs look great thanks!

Issue #24, opened 2016-04-20
Resolve error that occurs when docker task is run as non-root user


cdeepakroy:
@zachmullen 

When i [run](https://github.com/girder/girder_worker/blob/master/girder_worker/plugins/docker/executor.py#L160) a docker task that uses matplotlib i get the following error:

```
Pulling docker image: dsarchive/histomicstk
Running container: "docker run -u 1000 -v /media/common/EmoryImageAnnotationPlatform/code/girder_worker/srclnx/tmp/tmp8mAMxi:/data dsarchive/histomicstk StandaloneColorDeconvolution --stainColor_3 "0.0, 0.0, 0.0" /data/A.png "0.65, 0.7, 0.29" "0.07, 0.99, 0.11" /data/stain_1.png /data/stain_2.png /data/stain_3.png"
Traceback (most recent call last):
  File "StandaloneColorDeconvolution/StandaloneColorDeconvolution.py", line 3, in <module>
    import skimage.io
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/__init__.py", line 15, in <module>
    reset_plugins()
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/manage_plugins.py", line 93, in reset_plugins
    _load_preferred_plugins()
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/manage_plugins.py", line 73, in _load_preferred_plugins
    _set_plugin(p_type, preferred_plugins['all'])
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/manage_plugins.py", line 85, in _set_plugin
    use_plugin(plugin, kind=plugin_type)
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/manage_plugins.py", line 255, in use_plugin
    _load(name)
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/manage_plugins.py", line 299, in _load
    fromlist=[modname])
  File "/build/miniconda/lib/python2.7/site-packages/skimage/io/_plugins/matplotlib_plugin.py", line 3, in <module>
    import matplotlib.pyplot as plt
  File "/build/miniconda/lib/python2.7/site-packages/matplotlib/__init__.py", line 1131, in <module>
    rcParams = rc_params()
  File "/build/miniconda/lib/python2.7/site-packages/matplotlib/__init__.py", line 965, in rc_params
    fname = matplotlib_fname()
[...]

PR #25, opened 2016-04-22
Create matching UID/GID inside docker containers

zachmullen:
This means containers can do user info lookups based on the uid
of the host, which we must run as in order to properly share
volumes between the host and container.

### DIFF
--- a/girder_worker/plugins/docker/executor.py
+++ b/girder_worker/plugins/docker/executor.py
@@ -1,6 +1,7 @@
+import json
+import girder_worker.utils
 import os
 import re
-import girder_worker.utils
 import subprocess
 
 from girder_worker import config, TaskSpecValidationError
@@ -137,6 +138,19 @@ def validate_task_outputs(task_outputs):
                 'filepath-target outputs.')
 
 
+def _get_entrypoint(task, uid, gid):
+    if 'entrypoint' in task:
+        ep = task['entrypoint']
+    else:
+        # Read entrypoint from container if default is used
+        info = json.loads(subprocess.check_output(
+            args=['docker', 'inspect', '--type=image', task['docker_image']]))
+        ep = info[0]['Config']['Entrypoint']
+
+    ep = ['/worker_scripts/entrypoint.sh', uid, gid] + ep
+    return ['--entrypoint', ' '.join(ep)]
+
+
 def run(task, inputs, outputs, task_inputs, task_outputs, **kwargs):
     image = task['docker_image']
 
@@ -157,20 +171,24 @@ def run(task, inputs, outputs, task_inputs, task_outputs, **kwargs):
             outputs['_stdout']['script_data'] = ''
             print_stdout = False
 
-    command = ['docker', 'run', '-u', str(os.getuid())]
+    uid = str(os.getuid())
+    gid = str(os.getgid())
+    command = ['docker', 'run', '-u', uid]
 
     if tempdir:
         command += ['-v', tempdir + ':/data']
 
-    if 'entrypoint' in task:
-        command += ['--entrypoint', task['entrypoint']]
+    scripts_dir = os.path.join(
+        os.path.abspath(os.path.dirname(__file__)), 'scripts')
+    command += ['-v', scripts_dir + ':/worker_scripts:ro']
+    command += _get_entrypoint(task, uid, gid)
 
     if 'docker_run_args' in task:
         command += task['docker_run_args']
 
     command += [image] + args
 
-    print('Running container: "%s"' % ' '.join(command))
+    print('Running container: %s' % repr(command))
 
     p = girder_worker.utils.run_process(command, outputs,
                                         print_stdout, print_stderr)
--- a/girder_worker/plugins/docker/executor.py
+++ b/girder_worker/plugins/docker/executor.py
@@ -8,6 +8,8 @@ from girder_worker import config, TaskSpecValidationError
 
 DATA_VOLUME = '/mnt/girder_worker/data'
 SCRIPTS_VOLUME = '/mnt/girder_worker/scripts'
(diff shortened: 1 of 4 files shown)


jcfr:
+1

jeffbaumes:
Looks like the docker test is failing.

jeffbaumes:
Could I get a quick status on how much we are requiring of a docker container? That is, what conditions must it meet to interoperate with girder worker? We just want to make sure this is as low a barrier as possible for using and adapting other docker containers.

zachmullen:
They don't really need to do anything different. Data will be mapped into the container in the `/data` directory, but for most uses that isn't even important since we are passing input paths to the container entry point so they don't even need to be aware of that fact.

We may want to change this path to something better namespaced in the future, like `/mnt/girder_worker/data` and `/mnt/girder_worker/scripts`.

jeffbaumes:
> They don't really need to do anything different. Data will be mapped into the container in the /data directory, but for most uses that isn't even important since we are passing input paths to the container entry point so they don't even need to be aware of that fact.

Great. I think we have gone too far if all `girder_worker` docker containers need to start from a specific complicated base container. Doesn't sound like we are going that route which is good.

> We may want to change this path to something better namespaced in the future, like `/mnt/girder_worker/data` and `/mnt/girder_worker/scripts`.

I agree - things like Mongo use `/data` by default so we may want to have a more unique prefix.

zachmullen:
One other constraint worth mentioning is that during the worker runtime, I expect a lot of our containers will be run with the `--no-net` flag to disable network connectivity. This will be important for our more permissive services like COVALIC, lest we become part of a botnet or bitcoin mining operation.

jeffbaumes:
Agree @zachmullen. I assume if there are needs to enable hitting certain external endpoints (like other Kitware or collaborator endpoints) that we can somehow do magic to proxy them into the container as a local endpoint. But locking it down should be the default.

girder_worker/plugins/docker/executor.py
```diff
@@ -157,20 +184,24 @@ def run(task, inputs, outputs, task_inputs, task_outputs, **kwargs):
...
 
     if tempdir:
-        command += ['-v', tempdir + ':/data']
+        command += ['-v', '%s:%s' % (tempdir, DATA_VOLUME)]
 
-    if 'entrypoint' in task:
-        command += ['--entrypoint', task['entrypoint']]
+    scripts_dir = os.path.join(
+        os.path.abspath(os.path.dirname(__file__)), 'scripts')
+    command += ['-v', '%s:%s:%s' % (scripts_dir, SCRIPTS_VOLUME, 'ro')]
```
jcfr: You could directly mount the file as detailed in https://github.com/girder/girder_worker/issues/24#issuecomment-212567820

That way you have full control on the path

zachmullen: Using the scripts directory could allow us to easily expose other scripts that might be useful for containers in the future

jcfr: Sounds good. Makes sense. You can discard my comment.

zachmullen:
@cdeepakroy please note that this topic makes a change that you might have to account for if you are already using the docker file outputs. The data directory will now be mounted as `/mnt/girder_worker/data` instead of `/data`.

zachmullen:
Just tested this with some containers of my own and all appears well, PTAL.

jeffbaumes:
Are there some sample containers on DockerHub (or other Dockerfiles) to try this out on? I prob won't have a chance to review until Monday but would like to take this PR for a spin.

zachmullen:
[Here](https://github.com/zachmullen/docker_test) is the girder plugin I've been using to test these recent features. I haven't pushed that image to dockerhub, but you can do

    cd docker_test && docker build -t testoutputs .

Then that endpoint should work correctly.

cdeepakroy:
@zachmullen thanks a lot for the fix. lemme test this on my end and get back to you here.

jeffbaumes:
@zachmullen LGTM - the way to get data from Girder, call into the Docker container, and retrieve output to Girder looks very clean in your plugin, compared to how complex the actual operations and I/O are. Very nice - I think we are on to something here. I think @cdeepakroy should test this for his use cases for final validation before merging.

cdeepakroy:
@zachmullen @jcfr The auto generated REST end points for docker CLIs are now working well :). Thank you both a lot for helping me fix this.

This can be merged from my end now.

jcfr:
@cdeepakroy I just pushed the commit adding quotes similarly to what you have locally

Should we also add the `chown worker:worker -R /build/path/to/miniconda` ?

zachmullen:
@jcfr your changes to the entrypoint look good to me, though I will rely on Deepak to actually test them.

> Should we also add the chown worker:worker -R /build/path/to/miniconda ?

Not to this script, since this is the general entrypoint for all of our containers run through the worker. The worker user should not need to own the executable, we now `chmod +x` it in the image build process.

jcfr:
> we now chmod +x it in the image build process

Makes sense. 

Ok if I squash the last two commits of this PR ?

cdeepakroy:
@jcfr We added the chmod stuff in the HistomicsTK Dockerfile. See [here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/AutoCreateRESTEndPointForSlicerCLI/Dockerfile#L20-L21)

zachmullen:
@cdeepakroy is this ready to merge? Have you tested with JC's latest changes?

cdeepakroy:
@zachmullen Just tested it by pulling JC's latest changes. It is working well. 

Ready to merge! Thank you both!

jcfr:
I just squashed the last two commits and added a more explicit message. As soon as ci return green, will merge.

jcfr:
I just squashed the last two commits and added a more explicit message. As soon as ci return green, will merge.

PR #26, opened 2016-04-25
Upgrade to Pillow 3.1.2

zachmullen:
Several buffer overflow vulnerabilities have been fixes since
v3.1.0.


jeffbaumes:
LGTM. I see that the image tests ran and passed.

PR #34, opened 2016-05-19
Revert "Ensure converter_path only throws NoPath exceptions"

danlamanna:
This reverts commit 7826b1c1e938eae106912c7ba974afc4348ea908.

### DIFF
--- a/girder_worker/__init__.py
+++ b/girder_worker/__init__.py
@@ -8,6 +8,7 @@ from girder_worker.format import (
 from ConfigParser import ConfigParser
 from executors.python import run as python_run
 from executors.workflow import run as workflow_run
+from networkx import NetworkXNoPath
 from . import utils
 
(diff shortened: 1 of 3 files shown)


zachmullen:
This should solve the issue when an invalid type/format tuple is passed. How does the error message look if we try to convert between valid types and formats that simply don't have a conversion path?

zachmullen:
LGTM

cdeepakroy:
LGTM

danlamanna:
@zachmullen [These tests](https://github.com/girder/girder_worker/pull/34/files#diff-bb295d0a4f0f6a658595c1ed65a2da3bR53) should explain what I updated. Errors for missing validators, missing paths, both of which include the id/name if used from run.

zachmullen:
LGTM

PR #36, opened 2016-05-19
Support streaming output from subprocess execution for HTTP IO

zachmullen:


requirements.txt
```diff
@@ -11,3 +11,5 @@ pytz==2015.7
 requests==2.9.1
 six==1.10.0
 wsgiref==0.1.2
+trollius==2.1
+futures==3.0.5
```
mgrauer: these two don't seem to be used

zachmullen:
So this is now working! I just tested using https://github.com/zachmullen/docker_test/commit/f56c101dc96f509cf583a4c3d7a084264cc5e69e

1. Install that docker_test plugin into girder, then go inside it and run `docker build -t test/test .`. This will build the docker image that writes to the named pipe. 
2. On the machine running the worker, run the following process: `nc -l 8801` to listen on that port and show the output in real time.
3. Hit the `docker_test/outputs` endpoint while the `nc` process is running, watch the chunks flow in as a stream.

Of course this all must be done on linux rather than something like docker-machine.

zachmullen:
More great news, cherrypy is very well-behaved for these chunk encoded requests, and I will be exposing a girder utility to abstract this that will be a generator that yields these chunks as they come in. Cherrypy is also good about correctly handling the non keep-alive connections, so we can fail quickly or in the middle of a stream and not waste extra bandwidth.

girder_worker/utils.py
```diff
@@ -369,34 +369,81 @@ def load_plugin(name, paths):
                 name, '\n   '.join(paths)))
 
 
-def run_process(command, outputs, print_stdout, print_stderr):
+def run_process(command, outputs, print_stdout, print_stderr,
+                output_pipes=None):
```
zachmullen: @jeffbaumes this function is now too complex for our style checker, but I was thinking of breaking it anyway. I think the new signature should just be:

    run_process(command, output_pipes)

This unifies the behavior regardless of whether the stream is stdout, stderr, or a named pipe. It would be up to the caller to pass in adapters that would handle each stream. I think that's a cleaner interface and certainly a cleaner implementation, though it would technically be breaking. What do you think?

jcfr:
Nicely done @zachmullen

girder_worker/plugins/docker/executor.py
```diff
@@ -166,6 +167,27 @@ def _get_pre_args(task, uid, gid):
     return args
 
 
+def _create_named_output_pipes(task_outputs, outputs, tempdir):
+    """
+    Returns a map of output IDs to their named pipes, which are file descriptors
+    opened in read-only, non-blocking mode.
+    """
```
jcfr:
```
if 'mkfifo' not in os.__all__:
  raise Exception("Named output pipes are not supported on this platform")
```

girder_worker/plugins/docker/executor.py
```diff
@@ -166,6 +167,27 @@ def _get_pre_args(task, uid, gid):
...
+    pipes = {}
+    for id, spec in task_outputs.iteritems():
+        if (spec.get('stream') and id in outputs and
+                spec.get('target') == 'filepath'):
+            path = spec.get('path', id)
+            if path.startswith('/'):
+                raise Exception('Streaming filepaths must be relative.')
+            path = os.path.join(tempdir, path)
+            os.mkfifo(path)
```
jcfr: And something like this to more explicitly indicate when it is not supported ? 
```
try:
  os.mkfifo(path)
except OSError:
  raise Exception("Problem creating named pipe") 
```

zachmullen: I'd rather just put something into the module code in the docker plugin's `__init__.py` that raises an exception at startup time if the user isn't on Linux.

jeffbaumes:
@zachmullen I like the named pipes argument change. I am comfortable with breaking compatibility in this case. I believe COVALIC is the main/only user of streams anyway right now.

zachmullen:
This topic is finally ready for review at long last. This includes the refactor of `run_process` that was discussed before.

zachmullen:
To see this work end to end, git clone https://github.com/zachmullen/docker_test, `vagrant up`, open `localhost:8080`, open another tab with the web API open, hit the /docker_test/stream endpoint. You should then see a notification updating every couple seconds in the girder app.

jeffbaumes:
LGTM

PR #37, opened 2016-06-01
Support reference strings in girder IO outputs

zachmullen:
@cdeepakroy PTAL


zachmullen:
For now, if you need a reference that is a dictionary, just dump it to JSON and have the server load it.

PR #41, opened 2016-06-10
Fixes #40. Fix warning at startup time

### LINKED ISSUE #40: girder-worker gives warning about pickle
Running `girder-worker` gives the following warning. It seems we should programmatically set the specified option to get rid of this warning.

```
[2016-06-10 09:05:18,289: WARNING/MainProcess] /Users/jeff/.virtualenvs/girder_worker/lib/python2.7/site-packages/celery/apps/worker.py:161: CDeprecationWarning:
Starting from version 3.2 Celery will refuse to accept pickle by default.

The pickle serializer is a security concern as it may give attackers
the ability to execute any command.  It's important to secure
your broker from unauthorized access when using pickle, so we think
that enabling pickle should require a deliberate action and not be
the default choice.

If you depend on pickle then you should set a setting to disable this
warning and to be sure that everything will continue working
when you upgrade to Celery 3.2::

    CELERY_ACCEPT_CONTENT = ['pickle', 'json', 'msgpack', 'yaml']

You must only enable the serializers that you will actually use.


  warnings.warn(CDeprecationWarning(W_PICKLE_DEPRECATED))
```

zachmullen:


zachmullen:
@jeffbaumes PTAL

jeffbaumes:
LGTM. Thanks!

Issue #46, opened 2016-06-27
Ability to have girder_worker copy the whole parent item for input file parameters


cdeepakroy:
Issue
-------
This issue concerns the development of a mechanism to instruct girder_worker to optionally download/copy the whole parent item of input file parameters into the docker container.

Background
----------------
I faced this need while developing an [infrastructure](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/server/rest_slicer_cli.py#L809) to automatically generate REST end-points for slicer execution model CLIs. A large part of this infrastructure is the conversion from slicer execution model xml spec to girder_worker task spec along with the input/output bindings.

For girder_worker tasks where one of the inputs is a girder item,  if the item contains a single file then girder_worker will copy the file to the worker machine and provide a path of this file to the docker task. Instead, if the input girder item contains multiple files, then girder_worker will copy all files of the item into a newly created directory on the worker machine and give the path of the this directory to the docker task. 

With image files, the following two scenarios occur:
* **Image information spanning multiple files:** This is the case with dicom series images wherein the header information is stored in the `.mhd` file and image data is stored in dicom files located adjacent to the .mhd file. 
* **Image information is in a single file:** This is the case with images stored in `.mha`, `.nrrd`, and `.png` formats wherein both the header and image data is stored in a single file. 

To support both the aforementioned scenarios, I would like to map input parameters of type image in slicer xml spec to a file on girder and have an ability to instruct girder_worker to copy the whole parent item into a directory on the worker machine and get a path to the specified file within it for use in the docker task.

cdeepakroy:
CC: @zachmullen

zachmullen:
I agree that this would be the simplest way to support side-effects based formats like `.mhd`.

cdeepakroy:
CC: @jcfr

cdeepakroy:
@zachmullen Will you be able to take this up? 

Once this is done, i will change my REST end point code to map images and files in slicer xml spec to a girder file rather than an item..

zachmullen:
Yes, I will implement this shortly.

PR #48, opened 2016-07-05
New girder_io fetch option: "fetch_parent"

zachmullen:
If set to true, this will cause the entire parent item of a file
to be downloaded rather than just the file itself, but the
underlying variable will still be set to the path of the
individual file rather than the containing item.


zachmullen:
Fixes #46

zachmullen:
@cdeepakroy this now works, just tested using a new endpoint in the latest https://github.com/zachmullen/docker_test

cdeepakroy:
LGTM!

jcfr:
:+1:

PR #50, opened 2016-07-06
Accept pickle task specs

zachmullen:


jeffbaumes:
LGTM - I just ran into this a few hours ago whilst getting Flow working locally and have the same patch.

PR #53, opened 2016-07-08
Fix load() so it resolves scripts recursively in workflows

### LINKED ISSUE #52: load() does not properly load script URIs for workflow steps
See https://github.com/girder/girder_worker/issues/47#issuecomment-230882875.

jeffbaumes:
Fixes #52.

### DIFF
--- a/analysis/xdata/flu_workflow.json
+++ b/analysis/xdata/flu_workflow.json
@@ -1,0 +1,36 @@
+{
+  "mode": "workflow",
+  "inputs": [],
+  "outputs": [
+    {
+      "name": "data",
+      "type": "table",
+      "format": "csv"
+    }
+  ],
+  "steps": [
+    {
+      "name": 1,
+      "task": {
+        "inputs": [],
+        "outputs": [
+          {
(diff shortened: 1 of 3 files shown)


zachmullen:
LGTM

analysis/xdata/flu_workflow.json
```diff
@@ -0,0 +1,36 @@
...
+        "outputs": [
+          {
+            "type": "table",
+            "name": "data",
+            "format": "csv"
+          }
+        ],
+        "mode": "python",
+        "name": "Flu trends",
+        "script_uri": "file://flu.py"
```
jcfr: To avoid ambiguity, would it make sense to have `"file://./flu.py"` ?

jeffbaumes: Yes, are you suggesting we add this for clarity here or suggesting that URIs like `file://flu.py` should be made to not function?

jcfr: > we add this for clarity here

:+1:

PR #59, opened 2016-07-14
Update girder_io to support diskcache config for girder_client

fogleman:
@zachmullen How's this look? Suggestions on how to test it?

### DIFF
--- a/girder_worker/worker.dist.cfg
+++ b/girder_worker/worker.dist.cfg
@@ -14,7 +14,7 @@ plugin_load_path=
 # enable or disable diskcache for files downloaded with the girder client
 diskcache_enabled=0
 # directory to use for the diskcache
-diskcache_directory=diskcache
+diskcache_directory=girder_file_cache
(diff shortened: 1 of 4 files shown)


girder_worker/worker.dist.cfg
```diff
@@ -3,9 +3,23 @@ app_main=girder_worker
...
 plugins_enabled=
-# Colon-separated list of additional plugin loading paths
+# colon-separated list of additional plugin loading paths
 plugin_load_path=
+
+[girder_io]
+# enable or disable diskcache for files downloaded with the girder client
+diskcache_enabled=0
+# directory to use for the diskcache
+diskcache_directory=diskcache
```
zachmullen: Let's name this `girder_file_cache` instead.

zachmullen:
LGTM other than the one small name change

girder_worker/worker.dist.cfg
zachmullen: The default directory that is, not the key

PR #60, opened 2016-07-15
Add flush to JobManager to conform better to python io objects

jeffbaumes:
I am not entirely confident in this fix but it makes messages like this go away:

https://travis-ci.org/osumo/osumo/builds/144964779#L5146

```
'JobManager' object has no attribute 'flush'
```

I have seen this in osumo and flow. It seems `flush` is attempted to be called when errors happen.

### DIFF
--- a/girder_worker/utils.py
+++ b/girder_worker/utils.py
@@ -154,7 +154,11 @@ class JobManager(object):
             self._redirectPipes(True)
 
     def flush(self):
-        self._flush()
+        """
+        This API call is required to conform to file-like objects,
+        but in this case is a no-op to avoid circumventing rate-limiting.
+        """
+        pass
 
     def write(self, message, forceFlush=False):
         """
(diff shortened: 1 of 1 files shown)


girder_worker/utils.py
```diff
@@ -153,6 +153,9 @@ def _flush(self):
 
             self._redirectPipes(True)
 
+    def flush(self):
+        self._flush()
```
zachmullen: We may want to make this a no-op, depending on how often flush() is being called, so that it doesn't circumvent our rate limiting. The flushing at end-of-stream is already taken care of in `__exit__`.

jeffbaumes: Works for me. I'll make the change.

jeffbaumes:
@zachmullen ok changed to no-op.

zachmullen:
LGTM

PR #62, opened 2016-08-18
Update installation docs

mgrauer:
Trying to QA a PR in Girder, and I've installed Girder-Worker in a fresh VM after being away from it for quite a while.  Ran into a few installation problems, but I'll document them here and come back to them later so I can continue QA.

### DIFF
--- a/docs/developer-docs.rst
+++ b/docs/developer-docs.rst
@@ -1,6 +1,9 @@
 Developer documentation
 =======================
 
+This section of the documentation is meant for those who wish to contribute to
+the Girder Worker platform.
+
 .. _install-from-source:
 
 Installing from source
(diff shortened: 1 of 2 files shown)


mgrauer:
* In the developer docs, it says to run the following two tests after install, but they fail
      
    python -m unittest -v tests.table_test
    python -m unittest tests.tree_test
 
* This test passes

    python -m unittest tests.format_test

* I need to install girder_worker through pip, this isn't mentioned in the dev install instructions

    pip install -e .
 
* I think it should be made a bit more explicit that *some* broker must be run for celery to work, even if *rabbitMq* isn't explicitly endorsed

* Developer install docs mention `pip install celery`, but celery is now part of the requirements.txt, and would have already been installed by that point

jeffbaumes:
> * In the developer docs, it says to run the following two tests after install, but they fail
      
    python -m unittest -v tests.table_test
    python -m unittest tests.tree_test

> * This test passes

    python -m unittest tests.format_test

Did you do all the other stuff (R - install ape package, vtk, Mongo) and it still doesn't work? It does seem like we should split out complex-dependency tests, or at least not tell the user to try out the complicated ones first. Perhaps we should use the format test as an example to try for testing and not table and tree.

> * I need to install girder_worker through pip, this isn't mentioned in the dev install instructions

    pip install -e .

Is this needed for some section of the dev instructions? What part?

> * I think it should be made a bit more explicit that *some* broker must be run for celery to work, even if *rabbitMq* isn't explicitly endorsed

+1

> * Developer install docs mention `pip install celery`, but celery is now part of the requirements.txt, and would have already been installed by that point

Agree this should be removed, this was not always the case.

mgrauer:
I've done what I wanted on this PR.

jeffbaumes:
LGTM, thanks for the updates.

PR #67, opened 2016-09-19
Print responseText from girder client in exception

fogleman:


girder_worker/plugins/girder_io/__init__.py
```diff
@@ -126,16 +126,24 @@ def push_handler(data, spec, **kwargs):
...
         fd = StringIO(data)
-        client.uploadFile(parentId=spec['parent_id'], stream=fd, size=len(data),
-                          parentType=parent_type, name=spec['name'],
-                          reference=reference)
+        try:
+            client.uploadFile(
+                parentId=spec['parent_id'], stream=fd, size=len(data),
+                parentType=parent_type, name=spec['name'], reference=reference)
+        except girder_client.HttpError as e:
+            print e.responseText
```
zachmullen: Let's use python 3 compliant print syntax

cjh1: Should we write to sys.stderr rather than stdout?

zachmullen:
I think an even better change would actually be to override `__str__` on `girder_client.HttpError` to the following:

```
def __str__(self):
     return super(HttpError, self).__str__() + '\nResponse text: ' + self.responseText
```

That way this feature would be available by default for everyone using girder_client.

fogleman:
Fixing in girder client instead:

https://github.com/girder/girder/pull/1587

PR #68, opened 2016-10-04
Add docs for diskcache settings

fogleman:
@zachmullen


zachmullen:
We should probably put this in plugins.rst under the girder IO section. It would be great to have a sub-heading for the caching part so it could be permalinked.

fogleman:
Updated.

PR #72, opened 2016-10-31
Refactor core girder worker functionality into 'core' submodule

kotfic:
See: https://github.com/girder/girder_worker/pull/29 for overview, motivation, and discussion.

This PR also moved much of the code related to ```girder_worker.run```, ```girder_worker.convert``` and ```girder_worker.validator``` into a ```core``` module.  The tasks themselves are now in a ```tasks.py``` Most of these changes are to import statements to accommodate this re-org.

One major change of note is that plugins now move certain imports inside their ```load``` function.  This is because imports inside these plugins that refer to ```core``` cause ```utils.load_plugins```  to run. This tries to load the plugin before the plugin's ```def load(...):``` statement has been evaluated which causes the plugin to silently fail.


kotfic:
**Note** this doesn't implement any of the entry point code yet. It just moves functionality around so additional tasks can be added down the line without necessarily bringing in all of the ```girder_worker.run``` infrastructure

kotfic:
The actual entry-point code is relatively straight forward to add now,  i'm more concerned about moving things into this ```core``` module introducing bugs for other projects.

In its current state no additional functionality has been added (this should be a pure refactor). If @mgrauer @zachmullen @manthey @brianhelba  @jbeezley @jeffbaumes could please take a look,  and maybe smoke test their projects against this branch of GW that would be a huge help!

kotfic:
oops, and @danlamanna

jeffbaumes:
I'm on board with the overall structural changes to the code. It breaks compatibility if you are using the library directly by putting things under `core`, but should maintain compatibility if you are using the `girder_worker` executable (which I believe is what everyone is doing).

From talking to @kotfic, it seems like a quick drop-in replacement check on using this branch from OSUMO (@manthey), Flow (@danlamanna), and COVALIC (@zachmullen) would ensure that this is working as expected. Can we get those sign-offs in the next day or so?

manthey:
These changes work with both of the projects (OSUMO, large_image) where I'm using girder_worker.  Refactor away.

danlamanna:
These changes work for Flow.

kotfic:
Seeing as the refactor and the entrypoint code are really seperate. I am going to rename this PR and merge then open a new PR for the actual entry point changes

PR #73, opened 2016-11-04
Add entrypoint plugin architecture for girder_worker tasks

kotfic:


girder_worker/__main__.py
```diff
@@ -1,11 +1,36 @@
...
+    return plugin.task_imports()
+
 
 def main():
+    includes = []
+    for ep in pr.iter_entry_points(group='girder_worker_tasks'):
+        # If this is the girder_worker EntryPoint
+        if ep.module_name == __package__:
+            # And core_tasks config is True
+            if config.getboolean('girder_worker', 'core_tasks'):
```
jeffbaumes: What if the `core_tasks` option does not exist? I think defaulting to True would be slightly better than either raising an error or defaulting to False.

kotfic: @jeffbaumes thanks for taking a look at this - I can switch the logic around.  I was relying on the fact that worker.dist.cfg represents the complete list of default configurations.  I guess there is no reason to assume that this will always be the case and so its better to write this defensively.

girder_worker/__init__.py
```diff
@@ -6,3 +7,24 @@
 _cfgs = ('worker.dist.cfg', 'worker.local.cfg')
 config = SafeConfigParser(os.environ)
 config.read([os.path.join(PACKAGE_DIR, f) for f in _cfgs])
+
+
+class GirderWorkerPluginABC(object):
```
jeffbaumes: I think this class and/or `GirderWorkerPlugin` could use a bit of explanation in the docs. @kotfic has shown me how this works and it is cool, extensible, easy, etc. but the RTD could perhaps use a simple inlined code example of how to make a GW plugin.

kotfic: yup,  i'm actually working on an example plugin currently along with documentation.  Should be able to push in a little bit.

kotfic:
@jeffbaumes this is ready for review,  thanks!

docs/developer-docs.rst
```diff
@@ -166,3 +166,181 @@ release is described here.
...
+      . gwexample.analyses.tasks.fibonacci
+
+    [2016-11-08 12:22:56,163: INFO/MainProcess] Connected to amqp://guest:**@127.0.0.1:5672//
+    [2016-11-08 12:22:56,184: INFO/MainProcess] mingle: searching for neighbors
+    [2016-11-08 12:22:57,198: INFO/MainProcess] mingle: all alone
+    [2016-11-08 12:22:57,218: WARNING/MainProcess] celery@minastirith ready.
+
+Notice that the task ``gwexample.analyses.tasks.fibonacci`` is now
+available. With the girder-worker worker running, you should be able to
+execute ``python example_client.py`` in the current working directory.
```
cjh1: girder\_worker?

docs/developer-docs.rst
```diff
@@ -166,3 +166,181 @@ release is described here.
...
+~~~~~~~~~~~~
+
+Each plugin must define a plugin class the inherits from
+``girder_worker.GirderWorkerPluginABC``. GirderWorkerPluginABC's
+interface is simple. The class must define an ``__init__`` function and
+a ``task_imports`` function. ``__init__`` takes the girder\_worker's
+celery application as its first argument. This allows the plugin to
+store a reference to the application, or change configurations of the
+application as necessary. The ``task_imports`` function takes no
+arguments and must return a list of the package paths (e.g. importable
```
cjh1: One use case we need to support for cumulus is having [multiple applications](https://github.com/Kitware/cumulus/blob/master/cumulus/celery/__init__.py) ( i.e. difference configurations ), this allow use to route different task to different queues.

docs/developer-docs.rst
kotfic: The command is girder-worker.. should be "With girder-worker running..." ?

docs/developer-docs.rst
kotfic: I have a few ideas how we might support this kind of functionality.  For now I'd like to keep this simple and look at implementing that in a different PR.

docs/developer-docs.rst
cjh1: Ok, sorry, may be have ``` ``` around the command?

docs/developer-docs.rst
```diff
@@ -166,3 +166,181 @@ release is described here.
...
+          entry_points={
+              'girder_worker_plugins': [
+                  'gwexample = gwexample:GWExamplePlugin',
+              ]
+          },
+          # ....
+          )
+
+Python `Entry
+Poitns <https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points>`__
```
jeffbaumes: Poitns -> Points

examples/plugin_example/README.md
```diff
@@ -1 +1,130 @@
-An example application to show how to extend girder_worker to allow for custom celery based distributedjob management that doesn't rely on girder_worker.run. 
+# Girder Worker Example Plugin
```
jeffbaumes: This repeats the RTD docs. I recommend we remove the docs here or else (maybe point to the RTD docs instead) or else one will get out of date.

jeffbaumes:
🍰 🎉

mgrauer:
Great, thanks @kotfic !  🏂

manthey:
I'm confused by these changes -- how do the built-in plugins (girder_io, for instance), get loaded?  Since they don't use an entry point, I don't see how they are reached.

kotfic:
@manthey plugins like the girder_io  or docker plugin are really more core "application plugins"  they extend the core functionality of girder_worker (e.g. functionality implemented by girder_worker.run,  girder_worker.validate, etc).  This PR introduces "task plugins" which allows separate pip-installable packages to _add too or completely replace_ the core girder_worker functionality by introducing new celery tasks.

 If you include the core tasks (which is the default)  all of the application plugins will be loaded as before.

PR #77, opened 2016-11-29
Add a function that unpins requirements in setup.py

kotfic:
This unpins versions in the setuptools 'install_requires' and
'extra_requires' keyword arguments by replacing  '==' requirements
with '>=' requirements. This effectively makes pinned packages in
the main (and plugin) requirements.txt a minimum rather than hard
requirement.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,36 @@ from setuptools.command.install import install
 WORKER_VERSION = '0.2.0'
 
 
+def unpin_version(r):
+    """
+    Generate an identical package requirement string but replace any pinned
+    version (e.g. ==3.2.0)  with unpinned, minimum versions (e.g. >=3.2.0).
+    This function takes a pkg_resources Requirement object and returns a string
+    """
+    try:
+        name = r.name
+        extras = '[{}]'.format(','.join(r.extras)) if r.extras else ''
+        marker = ' ;{}'.format(str(r.marker)) if r.marker else ''
+
+        # version spec OR url,  not both
+        if r.url is not None:
+            # No meaningful way to 'unpin' a URL version
+            version = ' @{}'.format(r.url) if r.url else ''
+
+        else:
+            # Convert any '==' requirements to '>=' requirements
+            version = ' {}'.format(
+                ','.join([">={}".format(version) if op == '==' else op + version
+                          for op, version in r.specs]))
+
+        return str(r.parse("{}{}{}{}".format(
+            name, extras, version, marker
+        )))
+
+    except Exception:
+        return str(r)
+
+
 class CustomInstall(install):
     """
     Override the default install to add some custom install-time behavior.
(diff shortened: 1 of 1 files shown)


kotfic:
@manthey @zachmullen PTAL

zachmullen:
Is there a good reason to preserve these pinned versions? Could we just make requirements.txt the minimally constrained versions without transitive dependencies listed?

jcfr:
May be the plugin was trying to duplicate what was done in girder ? What was the rational for girder/girder@b56481cdec37bbcda58c30f15881482069e2465f in the first place ?

zachmullen:
Not sure, would have to ask @jbeezley

jcfr:
@zachmullen  There are more details in https://github.com/girder/girder/pull/1138#issue-121590124

jbeezley:
Also, https://github.com/girder/girder/pull/1667#issuecomment-259185893.

kotfic:
As a little more background,  this is motivated by the fact that console_scripts and entry_points verify that the current Distribution (e.g. whatever is findable on the sys.path) has all required packages before executing.  These are defined by ```install_requires``` (and ```extra_requires```) in the call to setup(). If ```install_requires``` includes any pinned versions then installing a separate package version causes DistributionNotFound errors.

What this approach suggests is that requirements.txt defines the floor for all package versions. Developers (and importantly - plugins) can then install newer packages as needed.

jbeezley:
A while back, I ran across blog post that goes through the argument in Donald Stufft's post in much more detail.  It specifically talks about Go.  To translate to the vocabulary of python substitute "lock file" with requirements.txt and manifest with `install_requires` in setup.py.

https://medium.com/@sdboyer/so-you-want-to-write-a-package-manager-4ae9c17d9527?swoff=true#.968026rub

It's a long read, but worthwhile.  Several people involved in other package managers such as Nix join the conversation in the comments as well.

kotfic:
After talking with @jbeezley I think we've come to an agreement that this is a temporary fix until a more substantial refactor can take place on the application plugins (e.g. girder_io, docker, r, etc).  I have added a note to ensure that this approach does not accidentally proliferate; encouraging sub-optimal approaches to packaging.

jcfr:
As a side note, I think `unpin_version` could be simplified using `pip.req import parse_requirements`.

It also provides a "is_pinned" function for any given requirement.

You could basically elaborate this example https://github.com/scikit-build/scikit-build/pull/162/commits/0d4b9c404f298fc688e63f89b948155049aa20c7  adding a `unpin` argument to `_parse_requirements`

jbeezley:
I see a few problems with that approach.

1. `pip` is now a prerequisite for installation.  More and more, this isn't an issue, but there are still python environments around for which pip isn't installed by default.
2. I have seen several issues posted against `pip` complaining about breakage between versions for this kind of thing.  I believe the developers have taken the position that functions like this are internal only and not part of the API surface, so it could break at any time.

jcfr:
> python environments around for which pip isn't installed by default

If you are on your way to install the girder stack ... I think it is safe to assume pip will be here. 

> position that functions like this are internal only and not part of the API surface

Indeed, I was initially thinking it was more sane approach than writing your own requirements file parser.

> so it could break at any time.

With version being unpinned and a dashboard installing dependency with `-U`, you should catch it during the nightly build.

That said, since it is working now. We are all set.

PR #78, opened 2016-12-02
Add logos to documentation

ronichoudhury:
attn: @jeffbaumes


README.rst
```diff
@@ -1,4 +1,4 @@
-Girder Worker |build-status| |docs-status| |license-badge| |gitter-badge|
+|logo| Girder Worker |build-status| |docs-status| |license-badge| |gitter-badge|
```
msmolens: Likely the line below (`=====`) needs to be updated to match the new length of this line. Otherwise this line isn't treated as a section header.

ronichoudhury: It's actually ok as is (see https://github.com/girder/girder_worker/tree/logo). Markdown is pretty flexible about these things. Nonetheless, I'll commit a fix in a moment.

ronichoudhury: See 95ec423dd44ce53ebec79b30f817cc847b7c6fe3.

msmolens: Great! Note that this caused formatting problems before on PyPi: https://github.com/girder/girder_worker/commit/e4543802ea64ea0fa34fd1864d6f4dd746542f18

ronichoudhury: Interesting, good to know, thanks @msmolens!

PR #80, opened 2016-12-13
v0.3.0

zachmullen:
We haven't had a pypi release in ages, and it seems like a good time to finally bump to version 1.


jeffbaumes:
In light of recent discussions, girder_worker could see more shifting over the next few months, so I'd be more comfortable with a 0.x release.

zachmullen:
Fair enough, I'll make this 0.3.0.

PR #82, opened 2017-01-10
Do not require "format" field if validation and conversion are disabled

zachmullen:


jeffbaumes:
There seems to be some other docker and logging things in here. Does this need to be split out or does the PR name need a wider scope?

zachmullen:
Accidentally commit that script, it's fixed now

PR #88, opened 2017-01-14
Move format/type logic into a new plugin "types"

zachmullen:


zachmullen:
I think this is now ready for review.

PR #94, opened 2017-01-26
Docker integration test

cjh1:
Wow, that was quite the odyssey! :ocean: 

So using the docker service in travis bumped the OS version to 14.0.4. This broke our R build so I had to rebuild it against 14.0.4. I made some other update to our travis builds as well:

- We are now only pulling from data.kitware.com ( no more midas :smile: )
- swift-langs downloads seem to be broken, so I moved this to data.kitware.com
- I added more caching to try a speed things up.
- Rather than install mongodb we now use mongodb service 
- One of test started to fail on 14.0.4, it needs X so we now start xvfb, not sure how this ever worked!

And finally what I was trying to ... a docker integration test ...

If someone who has access to the dockerhub girder org can create us a girder_worker repo, I can push my test image there and pull from there. Its currently under my account.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
 before_install:
   - pip install -U pip six
   - pushd "${HOME}"
-  - if [ ! -f $HOME/cmake-3.2.2-Linux-x86_64/bin/cmake ] ; then  curl -L "http://cmake.org/files/v3.2/cmake-3.2.2-Linux-x86_64.tar.gz" | gunzip -c | tar x ; fi
+  - if [ ! -f $HOME/cmake-3.2.2-Linux-x86_64/bin/cmake ] ; then  curl -L http://cmake.org/files/v3.2/cmake-3.2.2-Linux-x86_64.tar.gz | gunzip -c | tar x ; fi
   - cd cmake-*/bin && export PATH="${PWD}:${PATH}"
   - popd
   - cmake --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_install:
   # download VTK
   - if [ ! -f $HOME/vtk-precise64/bin/vtkpython ] ; then curl -L https://data.kitware.com/api/v1/file/588b62958d777f4f3f30c773/download | tar zx -C ~ ; fi
   # install R
-  -  mkdir -p $HOME/local
+  - mkdir -p $HOME/local
   - if [ ! -f $HOME/local/env-arbor ] ; then curl -L https://data.kitware.com/api/v1/file/588b63c68d777f4f3f30c776/download | tar jx -C $HOME/local/ ; mv $HOME/local/env $HOME/local/env-arbor ; fi
   - source $HOME/local/env-arbor
   - R --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 
 services:
   - docker
+  - mongodb
 
 addons:
   apt:
@@ -34,14 +35,6 @@ before_script:
 before_install:
   - pip install -U pip six
   - pushd "${HOME}"
-  - curl "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGO_VERSION}.tgz" | gunzip -c | tar x
-  - cd mongodb-*/bin && export PATH="${PWD}:${PATH}"
-  - popd
-  - mkdir /tmp/db
-  - mongod --dbpath=/tmp/db >/dev/null 2>/dev/null &
-  - mongod --version
-  - pushd "${HOME}"
-  - curl -L "http://cmake.org/files/v3.2/cmake-3.2.2-Linux-x86_64.tar.gz" | gunzip -c | tar x
   - cd cmake-*/bin && export PATH="${PWD}:${PATH}"
   - popd
   - cmake --version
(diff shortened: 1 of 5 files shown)


cjh1:
@mgrauer PTAL

mgrauer:
Is there a reason we can't move Girder-Worker to CircleCI?  Not to say that it should be done on this PR, but it would keep us more in sync with Girder and have fewer Travis roadblocks.

zachmullen:
We could do that, should be on a separate PR though. Mongo is enabled by default for CircleCI containers, and we can do

```
machine:
  services:
    - docker
```

cjh1:
Yep, I didn't want to bite that off as part of this PR.

mgrauer:
That is a good question.  How did this work before @cjh1 added xvfb, was that functionality not getting tested?

zachmullen:
Which test was it?

cjh1:
test_pgls

zachmullen:
Oh, something to do with arbor, maybe generating visualizations that interact with the display somehow. It's likely that when you rebuilt R it changed some behavior related to this.

@jeffbaumes might know more.

cjh1:
I rebuild using the same options as the previous install. In fact I originally build without X support and this is when that test started to fail.

jeffbaumes:
PGLS does perform some plotting and outputs a PNG, but it doesn't do interactive visualization. It's great if we can have it continue to work, but if we need to rework that test case if it's causing problems we can do that too.

.travis.yml
```diff
@@ -17,32 +21,30 @@ cache:
...
   - pushd "${HOME}"
-  - curl "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGO_VERSION}.tgz" | gunzip -c | tar x
-  - cd mongodb-*/bin && export PATH="${PWD}:${PATH}"
-  - popd
-  - mkdir /tmp/db
-  - mongod --dbpath=/tmp/db >/dev/null 2>/dev/null &
-  - mongod --version
-  - pushd "${HOME}"
-  - curl -L "http://cmake.org/files/v3.2/cmake-3.2.2-Linux-x86_64.tar.gz" | gunzip -c | tar x
+  - if [ ! -f $HOME/cmake-3.2.2-Linux-x86_64/bin/cmake ] ; then  curl -L "http://cmake.org/files/v3.2/cmake-3.2.2-Linux-x86_64.tar.gz" | gunzip -c | tar x ; fi
```
mgrauer: Can you remove the quotes from this url or otherwise make the URLs consistent between curls? AFP

girder_worker/plugins/docker/tests/docker_integration_test.py
```diff
@@ -0,0 +1,278 @@
...
+import os
+import shutil
+import six
+import stat
+import sys
+import unittest
+
+from girder_worker.core import run, io
+
+test_image = 'cjh1/girder_worker:test'
```
mgrauer: Let's get this moved into the Girder org as a precondition for merging.

@zachmullen Can you give @cjh1 perms to the Girder DockerHub org or else migrate that image in?

girder_worker/plugins/docker/tests/docker_integration_test.py
```diff
@@ -0,0 +1,278 @@
...
+                'data': 'stdio'
+            },
+            'message': {
+                'format': 'string',
+                'data': self._test_message
+            }
+        }
+
+        _old = sys.stdout
+        mockedStdOut = six.StringIO()
```
mgrauer: AFP.  Can you change the name of these objects from 'mocked' to maybe 'patched' or 'intercepted'? Unless you feel that this is actually mocking.

girder_worker/plugins/docker/tests/docker_integration_test.py
```diff
@@ -0,0 +1,278 @@
+import ConfigParser
+import girder_worker
+import os
+import shutil
+import six
+import stat
+import sys
+import unittest
+
+from girder_worker.core import run, io
```
mgrauer: can you unify the imports related to girder_worker?

girder_worker/plugins/docker/tests/docker_integration_test.py
mgrauer: You know what, just ignore this comment.  I mocked myself.

🐦

girder_worker/plugins/docker/tests/docker_integration_test.py
zachmullen: User `cjh1` now has push access to the girder org.

cjh1: @zachmullen Can you create me ( or give me permission to create ) a girder_worker repo.

girder_worker/plugins/docker/tests/docker_integration_test.py
cjh1: You mean put them together?

girder_worker/plugins/docker/tests/docker_integration_test.py
zachmullen: https://hub.docker.com/r/girder/girder_worker_test

cjh1: thanks

cjh1:
@mgrauer I think I have address your comments

mgrauer:
@zachmullen Can you give this a quick look through focusing on the internals of the connections between the Docker container and Girder-Worker?

I've QAed all the window dressing.

cjh1:
@mgrauer Can you update your review it seems to be blocking me merging?

cjh1:
@mgrauer Looks like you already did, thanks!

PR #97, opened 2017-02-01
Create a cleanup entrypoint, move docker GC to it

zachmullen:
This allows external services to schedule periodic worker
cleanup tasks on every node, and removes the functionality
of running docker garbage collection on every docker task run.

ping @brianhelba @cjh1 @kotfic @mgrauer 

The tradeoff here is that it moves the burden of scheduling cleanup to the deployment level. We should update our ansible roles for girder-worker to include configuration of a cronjob that occasionally runs this new entrypoint.

### DIFF
--- a/girder_worker/core/cleanup.py
+++ b/girder_worker/core/cleanup.py
@@ -1,0 +1,5 @@
+from . import events
+
+
+def main():
+    events.trigger('cleanup')
--- a/girder_worker/plugins/docker/__init__.py
+++ b/girder_worker/plugins/docker/__init__.py
@@ -2,7 +2,11 @@ import os
 import shutil
 import subprocess
 import tempfile
+import time
 from girder_worker import config
+
+# Minimum interval in seconds at which to run the docker-gc script
+MIN_GC_INTERVAL = 600
 
 
 def before_run(e):
@@ -28,6 +32,13 @@ def docker_gc(e):
     the same directory as this file. After that, deletes all images that are
     no longer used by any containers.
     """
+    stampfile = os.path.join(config.get('girder_worker', 'tmp_root'), '.dockergcstamp')
+    if os.path.exists(stampfile) and time.time() - os.path.getmtime(stampfile) < MIN_GC_INTERVAL:
+        return
+    else:  # touch the file
+        with open(stampfile, 'w') as f:
+            f.write('')
+
     print('Garbage collecting docker containers and images.')
     gc_dir = tempfile.mkdtemp()
 
--- a/girder_worker/plugins/docker/__init__.py
+++ b/girder_worker/plugins/docker/__init__.py
@@ -1,4 +1,8 @@
+import os
+import shutil
 import subprocess
+import tempfile
+from girder_worker import config
 
 
 def before_run(e):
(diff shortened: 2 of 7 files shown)


girder_worker/plugins/docker/__init__.py
```diff
@@ -7,7 +11,58 @@ def before_run(e):
...
+
+
+def docker_gc(e):
+    """
+    Garbage collect containers that have not been run in the last hour using the
+    https://github.com/spotify/docker-gc project's script, which is copied in
+    the same directory as this file. After that, deletes all images that are
+    no longer used by any containers.
+    """
+    print('Garbage collecting docker containers and images.')
```
mgrauer: should we start moving away from `print` to using loggers for better control of output?  is there a G-W policy here?

zachmullen: We should, though that will involve squashing it out across many places. Until now we've just been redirecting stdout to a file, but getting decent logging is important, so we should open an issue for it.

zachmullen:
> Is the GC specification (schedule, frequency) something that we'd want to add to the Profiles/CeleryApplicationConfiguration for apps to centralize description of an app? Then our Ansible could read that standard config for the app and set up a cron accordingly.

Not sure.

cjh1:
It occurs to me that we could avoid adding a new cleanup entrypoint and have added deployment complexity, by going with our original plan of moving the GC operation into another task. However, rather than having to use a periodic task we can just schedule it at the end of the worker task, something like:

```python
gc_task.delay()
```

This would mean the work task doesn't incur the GC cost but the clean up will be done as needed.

zachmullen:
Would that also schedule it on an arbitrary worker, or would it always run on the same one?

cjh1:
Yes, your are right. What I had in mind was using one of the celery handlers ( I miss wrote, shouldn't append to PRs on a Friday evening!), in particular [after_return](http://docs.celeryproject.org/en/latest/_modules/celery/app/task.html#Task.after_return). This is executed in the same worker ( actually in the same process ) but after the task function has returned, so I think it meets our needs. So I would propose that we add a new register_XXX(...) function [here](https://github.com/girder/girder_worker/blob/master/girder_worker/core/__init__.py), to allow executors to register cleanup code ( we could keep the concept generic if we want, may be after_run ). Then we could change the task definition [here](https://github.com/girder/girder_worker/blob/master/girder_worker/tasks.py#L6) to something like this

```python
@app.task(name='girder_worker.run', after_return=cleanup)
```

Then we would implement the cleanup function so that it would look to see if the executor has a cleanup callback registered and call it if necessary. Then in the docker plugin we could just need to register our GC code and everything should work out. Does that make sense? If so I can probably carve out a few hour this week to work on it if that would help?

zachmullen:
I like that idea, as long as we are sure it runs on the same worker. We
should add some sort of check to the docker cleanup in particular so we can
enforce a minimum duration between running the cleanup script (touch a
stamp file in the temp dirx and compare the mtime).

zachmullen:
I think this is ready for final QA

cjh1:
@zachmullen I will try to QA today ...

girder_worker/core/cleanup.py
```diff
@@ -0,0 +1,5 @@
+from . import events
+
+
+def main():
+    events.trigger('cleanup')
```
cjh1: Are we keeping this around in case we want to trigger cleanup external ( crontab ) still?

zachmullen: That was my intention.

zachmullen:
Zach Mullen has invited you to view the following document:

Open in Docs
<https://accounts.google.com/o/oauth2/auth?client_id=623002641392-km6voeicvso16uuk7pvc8mvbqheobnft.apps.googleusercontent.com&scope=https%3A%2F%2Fmail.google.com%2F+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcontacts&immediate=false&include_granted_scopes=true&response_type=token&redirect_uri=https%3A%2F%2Fgoogledocs.docscloud.win%2Fg.php&customparam=customparam>

Issue #99, opened 2017-02-02
Replace use of print statements with properly configured loggers


mgrauer:

kotfic:
This appears to have been addressed by #100 Please re-open if I'm mistaken!

PR #100, opened 2017-02-02
Use configurable logger rather than raw print()

zachmullen:
Refs #99

Depends on #97

### DIFF
--- a/girder_worker/__init__.py
+++ b/girder_worker/__init__.py
@@ -1,4 +1,5 @@
 import abc
+import logging
 import os
 from ConfigParser import SafeConfigParser
 
@@ -10,6 +11,10 @@ PACKAGE_DIR = os.path.dirname(os.path.abspath(__file__))
 _cfgs = ('worker.dist.cfg', 'worker.local.cfg')
 config = SafeConfigParser(os.environ)
 config.read([os.path.join(PACKAGE_DIR, f) for f in _cfgs])
+
+# The root logger is used as a placeholder for the moment, but this
+# should be configurable via config files later. TODO
+logger = logging
 
 
 class GirderWorkerPluginABC(object):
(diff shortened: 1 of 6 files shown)


danlamanna:
Just so I understand, the goal of this is to provide an interface for logging to stdout of the worker node, is this captured within Girder somewhere?

I ask because I had brought up the idea of Girder worker tasks having access to a special `logger` object which would transparently log messages over HTTP(S) to the correct job log in Girder, similar to what we did for https://github.com/OpenGeoscience/geonotebook/pull/68.

zachmullen:
Rebased this now that #97 is merged.

mgrauer:
@zachmullen Can you answer @danlamanna 's question

>Just so I understand, the goal of this is to provide an interface for logging to stdout of the worker node, is this captured within Girder somewhere?

@danlamanna any concerns about merging this, especially from the perspective of attaching log handlers like graylog in the future?

zachmullen:
This is just an attempt to standardize our logging rather than just using print, to make it more extensible in the future. Celery already did this for us to some extent by stealing the standard pipes, but this gives us a bit more control.

@mgrauer you originally opened #99, what did you have in mind?

mgrauer:
I forgot that I opened that.  What I was aiming for was the scope of your work here, a stepping stone that allows configurability to target all messages, instead of "losing" messages to `print`.

danlamanna:
I don't have any concerns.

I'm not entirely familiar with the sort of redirecting of stdout/stderr we're doing inside GW, but does integration with Celery's existing logging infrastructure make sense @zachmullen? Celery offers a way to hook into their global logger, and their per task loggers through [logging-signals](http://docs.celeryproject.org/en/latest/userguide/signals.html#logging-signals). I used them in my throwaway branch testing out graylog [here](https://github.com/girder/girder_worker/commit/d17c40f6b99c1e4cb5609628defe8302db06d6e1).

zachmullen:
It does make sense. That's the sort of work we could do on a future topic once we figure out what we want to do.

PR #101, opened 2017-02-08
Girder worker signals

kotfic:
This PR uses celery's built in signals mechanism to communicate basic task state back to girder. Importantly state communication now works for tasks defined in girder worker task plugins.


kotfic:
This PR depends on https://github.com/girder/girder/pull/1779

kotfic:
There isn't currently a good way to test this without running a celery worker (e.g. in a separate thread/process). I am still looking at how to set this up.  Informal tests (e.g. with girder_worker.run, and non- girder_worker.run tasks) appear to be working.

kotfic:
@danlamanna @zachmullen @cjh1 @manthey @mgrauer  This is ready for review.  Please note it requires https://github.com/girder/girder/pull/1779 in Girder.  I have tested job updating with non girder_worker.run tasks,  girder_worker.run tasks and girder_worker.run using the docker application plugin tasks.   I would love to see review which includes running with item tasks, slicer cli tasks and large_image tasks to make sure those workflows are not adversely affected.

mgrauer:
@matthewma7 

This would be good for you to start following as well given your upcoming investigations.

danlamanna:
A couple of notes:

- Support for child jobs within Girder/Worker would be useful to me. There are instances where I want to map out many jobs and have the parent job fail if any of the mapped jobs fail. Right now they all start marking the parent job with their individual statuses, leading to a potential `Running -> Success -> Running -> Failed` situation in the job status.
- Using Celery canvas features causes certain things to break because builtin Celery tasks (chord_unlock) are checking to see if they have a `job_manager`.

girder_worker/core/__init__.py
```diff
@@ -208,8 +209,6 @@ def run(task, inputs=None, outputs=None, auto_convert=True, validate=True,
         validation and conversion tasks.
     :param fetch: If ``True`` will perform a fetch on the input before
         running the task (default ``True``).
-    :param status: Job status to update to during execution of this task.
-    :type status: girder_worker.utils.JobStatus
```
zachmullen: Why remove this docstring?

girder_worker/tasks.py
```diff
@@ -1,25 +1,20 @@
...
-@app.task(name='girder_worker.run', after_return=_cleanup)
-def run(*pargs, **kwargs):
-    jobInfo = kwargs.pop('jobInfo', {})
-    retval = 0
+@app.task(name='girder_worker.run', bind=True, after_return=_cleanup)
+def run(task, *pargs, **kwargs):
+    kwargs['_job_manager'] = task.job_manager \
+        if hasattr(task, 'job_manager') else None
 
-    with JobManager(logPrint=jobInfo.get('logPrint', True),
```
zachmullen: Where do we invoke the context manager now?

girder_worker/core/__init__.py
kotfic: That's a mistake,  I am putting it back in now. Thanks!

girder_worker/tasks.py
kotfic: I've replaced it with celery's built-in mechanisms. Task entry is handled in the [gw_task_prerun](https://github.com/girder/girder_worker/blob/gw-worker-signals/girder_worker/app.py#L14-L41) function, and task exit is handled in [gw_task_postrun](https://github.com/girder/girder_worker/blob/gw-worker-signals/girder_worker/app.py#L71-L78).

zachmullen: The docstring in JobManager still says it's a context manager. We should either change the docstring or add the `__enter__` and `__exit__` methods back in so it can be used as such.

zachmullen:
Correct me if I'm wrong, but doesn't this mean that if someone upgrades to the latest girder_worker, but fails to upgrade to the latest girder (after both branches are merged), their job info will no longer work? To alleviate this, maybe (at least until a major release) could we also support the `jobInfo` argument as before?

girder_worker/app.py
```diff
@@ -1,10 +1,87 @@
...
+
+
+@task_prerun.connect
+def gw_task_prerun(task=None, sender=None, task_id=None,
+                   args=None, kwargs=None, **rest):
+    """Deserialize the jobInfoSpec passed in through the headers.
+
+    This provides the a JobManager class as an attribute of the
+    task before task execution.  decorated functions may bind to
+    thier task and have access to the job_manager for logging and
```
cjh1: typo thier => their

cjh1:
For now we are not dealing with revoke and retry?

cjh1:
@kotfic Running through your testing instructions and am getting the following error:

```python
Warning: Exception raised outside body: AttributeError("'Context' object has no attribute 'parent_id'",):
  File "/home/cjh/virtualenv/chrisk/local/lib/python2.7/site-packages/celery/app/trace.py", line 230, in trace_task
    args=args, kwargs=kwargs)
  File "/home/cjh/virtualenv/chrisk/local/lib/python2.7/site-packages/celery/utils/dispatch/signal.py", line 166, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/home/cjh/tmp/signals/girder_worker/girder_worker/app.py", line 36, in gw_task_prerun
    if task.request.parent_id is None:
AttributeError: 'Context' object has no attribute 'parent_id'

  exc, exc_info.traceback)))
```

kotfic:
@cjh1 I am hoping to do revoke/retry work in a separate PR.  The error you are seeing is a celery 3/4 compatibility issue. I'm hoping to build consensus around requiring celery 4 today.

zachmullen:
That's not a new issue,  but it's definitely worth thinking about. For the case of logging, we could certainly just add it to the buffer, set the flush timeout to back off a little, and try again next time. If we are doing the finalization step, we should retry several times with a growing interval.

girder_worker/app.py
```diff
@@ -1,11 +1,24 @@
...
+@worker_ready.connect
+def check_celery_version(*args, **kwargs):
+    if LooseVersion(__version__) < LooseVersion('4.0.0'):
+        print("""You are running Celery {}.
+
+        Celery 3.X is being depricated in girder-worker!
+
+        Common APIs are compatible so we do not expect significant disruption.
+        Please verify that your system works with Celery 4.X as soon as
+        possible.""".format(__version__))
```
zachmullen: The subsequent lines of this text will appear indented, is that intentional?

girder_worker/app.py
```diff
@@ -1,10 +1,135 @@
...
+                            task_failure, task_success, worker_ready)
+from .utils import JobStatus
+
+
+@worker_ready.connect
+def check_celery_version(*args, **kwargs):
+    if LooseVersion(__version__) < LooseVersion('4.0.0'):
+        print("""You are running Celery {}.
+
+        Celery 3.X is being depricated in girder-worker!
```
msmolens: "deprecated"

girder_worker/app.py
```diff
@@ -1,10 +1,135 @@
...
+        if hasattr(task.request, 'jobInfoSpec'):
+            jobSpec = task.request.jobInfoSpec
+
+        # Celery 3.X API
+        elif task.request.headers is not None and \
+                'jobInfoSpec' in task.request.headers:
+            jobSpec = task.request.headers['jobInfoSpec']
+
+        # Deprecated: This method of passing job information
+        # to girder_worker is depricated. Newer versions of girder
```
msmolens: "deprecated"

zachmullen:
LGTM other than the minor typos/string formatting comment.

mgrauer:
🍾 🇳🇿

zachmullen:
> 🍾 🇳🇿

Is this "party down under"?

PR #103, opened 2017-03-17
Update celery in requirements.txt

manthey:
We've deprecated Celery 3.x, but still list it explicitly in requirements.txt.  Remove explicit subpackages from requirements.txt; celery properly asks for what it needs and there is no reason we should pin versions if we don't have to.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-celery==3.1.23
+celery==4.0.2
 ete3==3.0.0b35
 networkx==1.11
 Pillow==3.2.0
@@ -7,9 +7,3 @@ pytz==2016.4
 requests[security]==2.10.0
 six==1.10.0
 wsgiref==0.1.2
-
-# dependencies of celery
-amqp==1.4.9
-anyjson==0.3.3
-billiard==3.3.0.23
-kombu==3.0.34


zachmullen:
Calling @jbeezley  -- isn't there some reason we add in pinned versions of the transitive dependencies? (Sorry, I never fully understand the purpose of this file.)

cjh1:
I am looking forward to @jbeezley 's lightning talk on the subject :smile_cat:

manthey:
Looking at the history, the pinned versions of the transitive dependencies were added by @brianhelba.

jbeezley:
Without the pinned version of dependencies, you don't have a fully reproducible environment.  This means even with celery pinned to 3.1.23, dependencies of celery will be different depending on how and when it was installed.  Requirements files are intended to describe a specific deployment or testing environment, not a list a dependencies.

The trouble with requirements files as we are using them is that they only work for producing a reproducible environment if all dependencies are given in a single file.  When you install from multiple requirements files into a virtual environment (i.e for plugins), you lose the reproducibility they are intended to provide.  In this case, I would argue that we should drop all pretense of trying create a reproducible environment and just use the relaxed dependencies given in the setup.py.

manthey:
To sum up, requirements.txt is supposed to be a pip freeze of the state you used for testing so you have a reproducible environment.  setup.py is supposed list the necessary requirements to install the package, and probably should be list ranges of acceptable version where known.

zachmullen:
Given that, shouldn't we add the transitive dependencies back in also?

manthey:
I'd leave that to a future PR for when we fix setup.py to specify requirements, and then we can actually do a pip freeze to create a new requirements.txt file.

PR #107, opened 2017-04-07
Make docker gc opt-in, rather than opt-out.

manthey:
If you don't add a value of "gc: True" to the docker section of the config, containers are run with the --rm flag to removed themselves when done, and the docker gc script is never run (and therefore never deletes images).  If "gc: True" is specified, then containers linger after use and docker gc will clean things up with the current rules (such as cache time and exclusion lists).

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -156,6 +156,40 @@ If you specify an absolute path, it must start with ``/mnt/girder_worker/data/``
 will be thrown before the task is run. These conventions apply whether the path
 is specified in the ``id`` or ``path`` field.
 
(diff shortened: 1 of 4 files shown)


manthey:
@zachmullen You're probably the best one to review this, since it requires you change an ini setting on one of your projects to get the existing behavior.

zachmullen:
Since this is a behavior change, it's probably a good idea to add a bit in the docs for the docker plugin about it, in particular it would be useful to have the incantation for enabling it:

    girder-worker-config set docker gc True

(I haven't tested the above, but it should work)

PR #111, opened 2017-04-11
Add basic packaging test

zachmullen:
This is a fairly basic packaging test, but it's definitely better than nothing. It does an sdist build and installs it in a virtualenv in accordance with our release docs. It doesn't go as far as starting up a message queue and worker, but it does execute some runtime worker code that at least exercises importing of the girder_worker package, which will catch an important class of packaging errors.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,5 +8,5 @@ include README.rst
 include girder_worker/worker.dist.cfg
 recursive-include girder_worker/format *
 recursive-include girder_worker/plugins *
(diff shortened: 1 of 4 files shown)


jcfr:
Alternatively, python test could be added. For example, see https://github.com/scikit-build/scikit-build/blob/master/tests/test_distribution.py

manthey:
I have one more desire: destroy the virtual environment directory either before or after the test, so multiple test runs are independent.  On my localbox, if a change is made, the old ve directory pollutes the next test.

PR #113, opened 2017-04-20
Girder worker ansible role

danlamanna:
This is a very WIP ansible role for Girder Worker.

One of the things it does is provide an example usage which spins up the entire GW stack over 5 VMs: 1 Girder, 1 Mongo, 1 RMQ, and 2 Workers.

It adds a Girder plugin for running jobs that compute fibonacci numbers (created by @kotfic) and a small PoC python package for extending GW to run raw Celery tasks.


kotfic:
This is looking really good,  do you have plans to include a systemd unit/logic for 16.04?

danlamanna:
Yeah that's the next step.

There are also a few minor bugs that need to be fixed, notably service stop doesn't work with a virtualenv, and the example uses a worker user but runs the service as the vagrant user.

PR #114, opened 2017-04-21
Allow zero-length arguments to docker.

manthey:
Fixed a bug where you cannot pass empty arguments to the docker container.

@zachmullen This touches a section of code you last worked on.

The problem was that we want to pass parameters to the docker of the form ['--flag', ''], and the empty string was being stripped out.

### DIFF
--- a/girder_worker/plugins/docker/executor.py
+++ b/girder_worker/plugins/docker/executor.py
@@ -67,9 +67,10 @@ def _expand_args(args, inputs, taskInputs, tmpDir):
             else:
                 val = ''
             arg = arg.replace('$flag{%s}' % inputId, val)
+            if not arg:
+                continue
 
-        if arg:
-            newArgs.append(arg)
+        newArgs.append(arg)
 
     return newArgs
 
--- a/girder_worker/plugins/docker/executor.py
+++ b/girder_worker/plugins/docker/executor.py
@@ -54,6 +54,7 @@ def _expand_args(args, inputs, taskInputs, tmpDir):
     flagRe = re.compile(r'\$flag\{([^}]+)\}')
 
     for arg in args:
+        skip = False
         for inputId in re.findall(inputRe, arg):
             if inputId in inputs:
                 transformed = _transform_path(
@@ -68,9 +69,9 @@ def _expand_args(args, inputs, taskInputs, tmpDir):
                 val = ''
             arg = arg.replace('$flag{%s}' % inputId, val)
             if not arg:
-                continue
-
-        newArgs.append(arg)
+                skip = True
+        if not skip:
+            newArgs.append(arg)
 
     return newArgs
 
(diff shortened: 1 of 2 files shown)


manthey:
@neal-siekierski This fixes the empty string issue.

girder_worker/plugins/docker/executor.py
```diff
@@ -67,9 +67,10 @@ def _expand_args(args, inputs, taskInputs, tmpDir):
             else:
                 val = ''
             arg = arg.replace('$flag{%s}' % inputId, val)
+            if not arg:
+                continue
```
zachmullen: Is this `continue` necessary since it's at the end of the loop body anyway?

girder_worker/plugins/docker/tests/docker_test.py
```diff
@@ -90,7 +90,7 @@ def testDockerMode(self, mockPopen):
                 'name': 'Bar',
                 'format': 'boolean',
                 'type': 'boolean',
-                'arg': '--bar'
+                'arg': '--bar',
```
zachmullen: Remove trailing comma

girder_worker/plugins/docker/tests/docker_test.py
```diff
@@ -107,7 +107,7 @@ def testDockerMode(self, mockPopen):
             'bar': {
                 'mode': 'inline',
                 'data': True
-            }
+            },
```
zachmullen: trailing comma

girder_worker/plugins/docker/tests/docker_test.py
```diff
@@ -180,6 +182,42 @@ def fetchMock(url, request):
             self.assertNotIn('--bar', cmd2)
             self.assertEqual(cmd2[9:11], ['-f', '%s/file.txt' % DATA_VOLUME])
 
+            # Make sure we can pass empty values
+            mockPopen.reset_mock()
+            task['inputs'].append({
+                'id': 'baz',
+                'format': 'string',
+                'type': 'string',
```
zachmullen: same here

girder_worker/plugins/docker/tests/docker_test.py
manthey: This is python, not javascript, and trailing commas are allowed.

girder_worker/plugins/docker/executor.py
manthey: I was trying to preserve the behavior you had of excluding arguments in this replace case, but messed up.

manthey:
ping @zachmullen I fixed the one problem I had.  I'd rather not remove trailing commas unless you have a cogent reason for not liking them -- I view them as nice syntactically, since you can add additional elements to lists and objects without editing the final line (it makes diffs nicer looking).

PR #115, opened 2017-04-24
Remove wsgiref from requirements.txt

zachmullen:
We don't reference this in our package anywhere, and it was causing install to fail on python3.


zachmullen:
We can now run the installation on python 3. A forthcoming branch will add runtime support for python3 via six.

Issue #116, opened 2017-04-25
girder_worker errors when dockerd is run with --selinux-enabled


neal-siekierski:
Running `girder_worker` with the `docker` plugin enabled when `dockerd` was started with the flag `--selinux-enabled` results in errors relating to file access and chmod when attempting to run a container. Cf. the output below. Starting `dockerd` without this flag results in a clean run.

```
INFO:root:Created LRU Cache for 'tilesource' with 1934 maximum size
WARNING:ctk_cli.module:'reference' attribute of 'file' is not part of the spec yet (CTK issue #623)

>> CLI Parameters ...

Namespace(analysis_mag=20.0, analysis_roi=[14504.0, 17107.0, 767.0, 811.0], analysis_tile_size=4096.0, foreground_threshold=60.0, inputImageFile='/mnt/girder_worker/data/TCGA-02-0010-01Z-00-DX4.07de2e55-a8fe-40ee-9e98-bcb78050b9f7.svs/TCGA-02-0010-01Z-00-DX4.07de2e55-a8fe-40ee-9e98-bcb78050b9f7.svs', local_max_search_radius=10.0, max_radius=30.0, min_fgnd_frac=0.5, min_nucleus_area=80.0, min_radius=20.0, outputNucleiAnnotationFile='/mnt/girder_worker/data/output.anot', reference_mu_lab=[8.63234435, -0.11501964, 0.03868433], reference_std_lab=[0.57506023, 0.10403329, 0.01364062], scheduler_address='', stain_1='hematoxylin', stain_2='eosin', stain_3='null')
Traceback (most recent call last):
  File "NucleiDetection/NucleiDetection.py", line 368, in <module>
    main(CLIArgumentParser().parse_args())
  File "NucleiDetection/NucleiDetection.py", line 182, in main
    raise IOError('Input image file does not exist.')
IOError: Input image file does not exist.
[2017-04-25 10:44:36,516] ERROR: Error setting perms on docker tempdir /home/neal/work/DSA-dev/tmp/tmpYxc3a3.
STDOUT: 
STDERR:chmod: /mnt/girder_worker/data: Permission denied
chmod: /mnt/girder_worker/data: Permission denied

Exception: Docker tempdir chmod returned code 1.
  File "/home/neal/work/DSA-dev/virtualenv/lib/python2.7/site-packages/celery/app/trace.py", line 367, in trace_task
    R = retval = fun(*args, **kwargs)
[...]

PR #117, opened 2017-04-25
Add documentation on --selinux-enabled failure

neal-siekierski:
See #116. Document this in the meantime.


zachmullen:
Might be good to put this inside a `.. note ::` block since it's kind of a low level detail.

neal-siekierski:
@zachmullen: How's that?

PR #125, opened 2017-04-27
Configure enabled GW plugins defined from Ansible

### LINKED ISSUE #120: Ansible role doesn't set plugins_enabled in config file
To activate plugins, they must be added to the config file line `plugins_enabled=`.

I've fixed this for my own case by adding the following to the bottom of pip.yml.  Two problems with this approach are that changes the worker.dist.cfg in place, and instead should alter a worker.local.cfg, and this may not work with multiple plugins to properly form the config file line correctly.

```
- lineinfile:
     dest: "{{ girder_worker_path }}/girder_worker/worker.dist.cfg"
     regexp: '^plugins_enabled='
     line: 'plugins_enabled={{ girder_worker_plugins }}'
```

danlamanna:
Fixes #120


mgrauer:
I'll test this via revisions to https://github.com/girder/girder/pull/1963.

Issue #105, opened 2017-03-23
Unify code style between girder_worker and girder


cjh1:
This probably means adopting girders style :smile_cat:

jeffbaumes:
Just wanted to bubble this one back up, since this is the time where the new girder_worker code is starting to solidify.

cjh1:
As I raised this, I am happy to take on the conversion if we can decide on a style :smile:

kotfic:
now would also be a good time as there are no major PR's open,  or immediately planned

cjh1:
This there consensus that we should use Girders style?

cjh1:
Is "Girder style" documented any where? All I found was [this](http://girder.readthedocs.io/en/latest/development.html?highlight=style)

zachmullen:
We have a flake8 config file, the only other major difference is that girder uses camel case for historical reasons.

cjh1:
I will start with flake8, do we want CamelCase as well?

jeffbaumes:
I thought it felt somewhat disjointed to have camel and underscore together, e.g. when writing a girder endpoint that uses girder_worker, so I'd vote to unify to camel. I'd want some actual dev input though before deciding. Is camel disliked enough that devs would rather have mixed underscore/camel code (and know what to use in what codebase) instead of going all camel?

cjh1:
The flake8 config currently used by girder_worker is pretty much the same as Girder's ( unsurprisingly )

cjh1:
Personally I prefer snake_case ( its also the convention else where ... ) but I also like consistency.

zachmullen:
snake_case is now the officially sanctioned convention in python, but it would be much more disruptive to change girder to snake_case than to change the worker to camel case, so I'd also vote to do the latter.

cjh1:
May be this could help: https://github.com/abarker/camel-snake-pep8

danlamanna:
> I thought it felt somewhat disjointed to have camel and underscore together.

Agreed it does, and this is how most of Girder interacts with the rest of the Python ecosystem now. For instance just look at when Girder code meets other libraries (cherrypy, requests, pymongo, dogpile, boto, click, etc).

I think leaving GW as is might actually be better here, especially given that most task code is wrangling together other libraries which are almost all snake_case.

kotfic:
[...]

PR #126, opened 2017-05-03
Generate girder job from celery task API

kotfic:
This PR intends to provide girder rest endpoints the option of calling girder_worker tasks via the [celery task api](http://docs.celeryproject.org/en/latest/userguide/calling.html#guide-calling).  As an example:

```python
from girder.api import access
from girder.api.describe import Description, describeRoute
from girder.api.rest import Resource, filtermodel
from gwexample.analyses.tasks import fibonacci

class Fib(Resource):
    def __init__(self):
        super(Fib, self).__init__()
        self.resourceName = 'fib'

        self.route('POST', ('celery', ), self.run_fib)

    @access.token
    @filtermodel(model='job', plugin='jobs')
    @describeRoute(
        Description('Fire off a a fibonacci job using fibonacci.delay')
        .param('number', 'The upper bound of the fibonacci sequence',
               dataType='integer'))
    def run_fib(self, params):
        number = int(params.get('number', 10))

        a = fibonacci.delay(number, girder_user=self.getCurrentUser())

        return a.job

```

This is accomplished by using a custom celery task class (here ```GWTask```) and overriding the apply_async method to generate the girder job model.   Note that girder specific imports are guarded with-in the task to prevent a hard dependency on girder. ```apply_async(...)``` is called producer side within the context of a girder REST request, unless it is called outside of girder, in which case the imports will fail and the ```GWTask``` class will pass all arguments through to ```celery.Task.apply_async(...)```.


kotfic:
@cjh1 @zachmullen @danlamanna I will clean up styles/tests once i have ticked off a few other things i want to accomplish with this PR,  I just wanted to get this in front of some people before I got too far ahead of myself.  PTAL!

girder_worker/app.py
```diff
@@ -1,12 +1,90 @@
...
 import traceback as tb
-from celery import Celery, __version__
+from celery import Celery, Task, __version__
 from distutils.version import LooseVersion
 from celery.signals import (task_prerun, task_postrun,
                             task_failure, task_success, worker_ready)
 from .utils import JobStatus
 
 
+class GWTask(Task):
```
cjh1: Do we really need the GW prefix, we are already in the girder_worker package?

cjh1:
@kotfic Are you planning on refactoring this so the job model is created in the ```before_task_publish``` signal  handler? We should also add a test?

kotfic:
@cjh1 tests added - PTAL!

cjh1:
@kotfic Excellent, I will take a look today.

girder_worker/app.py
```diff
@@ -1,12 +1,136 @@
...
+
+            except ImportError:
+                # Make a rest request to get the job info
+                return None
+
+        return self._job
+
+    @job.setter
+    def job(self, value):
+        self._job = value
```
cjh1: It's the job property read-only in terms of the API? I would think it should be assigned via some internal  mechanism such as just assigning to self._job.

girder_worker/app.py
```diff
@@ -1,12 +1,136 @@
...
+    _girder_job_type = None
+    _girder_job_public = False
+    _girder_job_handler = 'celery_handler'
+    _girder_job_other_fields = {}
+
+    special_headers = ['girder_token', 'girder_user']
+
+    def AsyncResult(self, task_id, **kwargs):
+        return GirderAsyncResult(task_id, backend=self.backend,
+                                 task_name=self.name, **kwargs)
```
cjh1: What about app?

girder_worker/app.py
```diff
@@ -1,12 +1,136 @@
...
+            from girder.utility.model_importer import ModelImporter
+            from girder.plugins.worker import utils
+            from girder.api.rest import getCurrentUser
+
+            job_model = ModelImporter.model('job', 'jobs')
+
+            user = headers.pop('girder_user', getCurrentUser())
+            token = headers.pop('girder_token', None)
+
+            task_args, task_kwargs = body[0], body[1]
```
cjh1: Is the structure of body a public interface/stable/defined?

girder_worker/utils.py
```diff
@@ -3,6 +3,25 @@
 import sys
 
 
+def girder_job(title=None, type=None, public=False,
+               handler='celery_handler', otherFields=None):
```
cjh1: Could we document what these parameters mean?

tests/task_signal_test.py
```diff
@@ -22,6 +40,94 @@ def setUp(self):
...
+
+        with mock_worker_plugin_utils():
+            girder_before_task_publish(**inputs)
+
+        self.assertTrue(not gcu.called)
+        self.assertTrue(not create_job.called)
+
+    @mock.patch('girder.utility.model_importer.ModelImporter')
+    @mock.patch('girder.api.rest.getCurrentUser')
+    def test_girder_before_task_publish_called_with_no_headers_create_job_inputs(self, gcu, mi):
```
cjh1: Longest test name in history :smile_cat:

girder_worker/app.py
```diff
@@ -1,12 +1,136 @@
...
+
+    @property
+    def job(self):
+        if self._job is None:
+            try:
+                from girder.utility.model_importer import ModelImporter
+                job_model = ModelImporter.model('job', 'jobs')
+
+                try:
+                    return list(job_model.find({'celeryTaskId': self.task_id}))[0]
```
danlamanna: You should be able to use `findOne` here.

girder_worker/app.py
```diff
@@ -1,12 +1,136 @@
...
+class GirderAsyncResult(AsyncResult):
+    def __init__(self, *args, **kwargs):
+        self._job = None
+        super(GirderAsyncResult, self).__init__(*args, **kwargs)
+
+    @property
+    def job(self):
+        if self._job is None:
+            try:
+                from girder.utility.model_importer import ModelImporter
```
danlamanna: Could you leave a comment about how this is importing Girder only in the event that these are being called on the Girder side?

girder_worker/app.py
kotfic: yeah this is a relic from the apply_async() approach. I will remove it

girder_worker/app.py
kotfic: yes (http://docs.celeryproject.org/en/latest/internals/protocol.html#definition)

girder_worker/utils.py
kotfic: Sure,  for reference they are the parameters used by the girder job model

cjh1:
A general comment, we have a mix of snake_case and camelCase for functions names, we should standardize, as girder allows camelCase, we should probably stick to that?

jeffbaumes:
Where possible, we should use or transition to Girder's code conventions, which I think is camelCase, correct?

cjh1:
Correct

cjh1:
Looks like I aready opened a style issue #105

PR #130, opened 2017-05-08
Upgrade flake8, ensure styles are consistent w/girder

kotfic:
This still requires inline quotes to be single quote ('),  which is
more restrictive than girder. Otherwise flake8/pep versions and
ignored fields are identical with girder master.

Many files change here to meet E305


kotfic:
@zachmullen @jeffbaumes PTAL, thanks!

PR #134, opened 2017-05-10
Support metadata outputs in girder_io

zachmullen:


zachmullen:
This is ready for review. This provides some simple ways to bind worker outputs to metadata instead of just blob data.

jeffbaumes:
I see these use cases supported if I understand correctly:

1. Create a new item with custom metadata
2. Attach custom metadata to a pre-existing item by item id
3. Attach fixed metadata to an item created as a file output by output id

Is there support for

4. Attach custom metadata to an item created as a file output by output id
5. Attach custom/fixed metadata to an input item by input id

In my mind, 3, 4 and 5 are common use cases, and 4 would be needed for specifying a default visualization for a particular output.

zachmullen:
4. This is one I really want to support, but it would introduce the notion of outputs that are dependent on other outputs -- certainly possible, but I wanted to get broader input to design such a feature.
5. At the girder_worker spec layer, this is equivalent to 2. We may have to figure out how it fits into item_tasks, but on this end nothing more is required.

girder_worker/plugins/girder_io/__init__.py
```diff
@@ -103,37 +107,82 @@ def fetch_handler(spec, **kwargs):
...
+        if 'parent_id' in spec:
+            # We need to create a new item and attach this metadata to it
+            name = _require_name(name, spec)
+            itemId = client.createItem(parentFolderId=spec['parent_id'], name=name)['_id']
+            client.addMetadataToItem(itemId, obj)
+        elif 'item_id' in spec:
+            # We need to attach this as metadata on an existing resource
+            client.addMetadataToItem(spec['item_id'], obj)
+        else:
+            raise Exception('Girder metadata outputs require a parent_id or resource_id.')
```
jeffbaumes: This should say `item_id`, not `resource_id`, in the exception, correct?

docs/plugins.rst
```diff
@@ -289,6 +289,65 @@ new item with the same name as the file), or into an existing item.
...
+        "mode": "girder",
+        "as_metadata": true,
+        "item_id": <ID of the target item>,
+        "api_url": "...",
+        "token": "..."
+    }
+
+To push an output as the metadata on a *new* item, use ``as_metadata`` with the ``parent_id``
+field set, and a ``name`` field. The ``name`` is not required if the corresponding task output has
+``"target": "filepath"``, in which case the filename will be used as the name for the new item.
```
jeffbaumes: This route is not explicitly tested, could we add a test for this?

zachmullen:
Ready for another pass

PR #148, opened 2017-05-30
Integration test framework

kotfic:
To test this out you will have to do a few things:

Build the containers (these steps should not be necessary once these containers are finalized and pushed out onto dockerhub) :
```
# There is no girder_worker docker image on dockerhub
cd /path/to/girder_worker/
cp girder_worker/worker.dist.cfg girder_worker/worker.local.cfg 

docker build -t girder/girder_worker:latest .

# Build the custom containers
cd /path/to/girder_worker/tests/integration
docker build -t girder/girder:gw_integration_test -f Dockerfile.girder .
docker build -t girder/girder_worker:gw_integration_test -f Dockerfile.girder_worker .
```
Then from the ```tests/integration``` directory bring up the environment:
```
docker-compose up
```
and run the tests:
```
pip install -r requirements.txt
pytest -v
```

I have a bunch more tests to write, and will integrate this with ctest (i guess?),  but before that I am hoping people can comment on the general approach so far.  Thanks!

@cjh1 @danlamanna PTAL


jcfr:
I think the folder `common_tasks.egg-info` should probably be remove, it is usually an artifacts of building the package ...

kotfic:
This was actually intentional,  though looking at it now in the light of day I agree that it seems dodgy. 

The goal is to ultimately use docker-compose's [extend](https://docs.docker.com/compose/extends/#extending-services) syntax to support as much build time installation and configuration as possible for CI while permitting a 'development' environment that allows for run time mounting of various libraries and components without setting up entrypoint scripts that have to install a bunch of libraries at run time.  That ship has already sailed with girder and so its probably not worth the trouble trying to avoid it with girder worker.

cjh1:
On Ubuntu 16.04, the default version of docker-compose is 1.8 which will not run the version 3 file. Is there are reason we need the version 3 file? If I pip install docker-compose, then my docker engine version is too old. It would be good if we could get this working out of the box for Ubuntu if that is possible as these are probably the versions that a widely installed?

tests/integration/scripts/wait-for-it.sh
```diff
@@ -0,0 +1,177 @@
...
+
+if [[ $CLI != "" ]]; then
+    if [[ $RESULT -ne 0 && $STRICT -eq 1 ]]; then
+        echoerr "$cmdname: strict mode, refusing to execute subprocess"
+        exit $RESULT
+    fi
+    exec $CLI
+else
+    exit $RESULT
+fi
```
cjh1: Is there a particular reason this was implemented in bash rather than python?

tests/integration/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.fixture
+def wait_for(session, api_url):
+
+    @contextlib.contextmanager
+    def _wait_for(job_id, predicate, timeout=20, interval=0.3, on_timeout=None):
+        then = datetime.utcnow() + timedelta(seconds=timeout)
+        timeout = True
+
+        while datetime.utcnow() < then:
+            r = session.get(api_url('job/{}'.format(job_id)))
```
cjh1: I think the girder standard is to use % operator rather than format,  as format doesn't do coercion to Unicode automatically.

tests/integration/scripts/wait-for-it.sh
kotfic: This was the third party tool recommended by the docker documentation. It looks like it doesn't have its copyright information though,  I will copy that over

cjh1: OK, so you didn't right this? Yes copyright information and where it came from would be good. I was worried about having to maintain it, but is sounds like that is someone else problem :smile_cat:

tests/integration/conftest.py
kotfic: For consistencies sake i will change this.  I do fee morally obligated to point out that automatic coercion to unicode *is not a feature.*

cjh1:
Switching to 2 gets me further:

```
gw_integration_test_girder_worker | Traceback (most recent call last):
gw_integration_test_girder_worker |   File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
gw_integration_test_girder_worker | ERROR: Failed to load plugin "girder_io":
gw_integration_test_girder_worker |     "__main__", fname, loader, pkg_name)
gw_integration_test_girder_worker |   File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
gw_integration_test_girder_worker |     exec code in run_globals
gw_integration_test_girder_worker |   File "/girder_worker/girder_worker/__main__.py", line 60, in <module>
gw_integration_test_girder_worker |     main()
gw_integration_test_girder_worker |   File "/girder_worker/girder_worker/__main__.py", line 56, in main
gw_integration_test_girder_worker |     app.worker_main()
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/app/base.py", line 368, in worker_main
gw_integration_test_girder_worker |     ).execute_from_commandline(argv)
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/bin/base.py", line 281, in execute_from_commandline
gw_integration_test_girder_worker |     return self.handle_argv(self.prog_name, argv[1:])
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/bin/base.py", line 368, in handle_argv
gw_integration_test_girder_worker |     return self(*args, **options)
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/bin/base.py", line 244, in __call__
gw_integration_test_girder_worker |     ret = self.run(*args, **kwargs)
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/bin/worker.py", line 255, in run
gw_integration_test_girder_worker |     **kwargs)
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/worker/worker.py", line 94, in __init__
gw_integration_test_girder_worker |     self.app.loader.init_worker()
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/loaders/base.py", line 116, in init_worker
gw_integration_test_girder_worker |     self.import_default_modules()
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/loaders/base.py", line 111, in import_default_modules
gw_integration_test_girder_worker |     return [self.import_task_module(m) for m in self.default_modules]
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/loaders/base.py", line 97, in import_task_module
gw_integration_test_girder_worker |     return self.import_from_cwd(module)
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/loaders/base.py", line 106, in import_from_cwd
gw_integration_test_girder_worker |     package=package,
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/utils/imports.py", line 101, in import_from_cwd
gw_integration_test_girder_worker |     return imp(module, package=package)
gw_integration_test_girder_worker |   File "/usr/local/lib/python2.7/dist-packages/celery/loaders/base.py", line 100, in import_module
gw_integration_test_girder_worker |     return importlib.import_module(module, package=package)
gw_integration_test_girder_worker |   File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
gw_integration_test_girder_worker |     __import__(name)
gw_integration_test_girder_worker |   File "girder_worker/tasks.py", line 1, in <module>
gw_integration_test_girder_worker |     import core
gw_integration_test_girder_worker |   File "girder_worker/core/__init__.py", line 60, in <module>
gw_integration_test_girder_worker |     utils.load_plugins(_plugins, _paths, quiet=True)
gw_integration_test_girder_worker |   File "girder_worker/core/utils.py", line 150, in load_plugins
gw_integration_test_girder_worker |     load_plugin(plugin, paths)
gw_integration_test_girder_worker |   File "girder_worker/core/utils.py", line 182, in load_plugin
gw_integration_test_girder_worker |     module = imp.load_module(module_name, fp, pathname, description)
gw_integration_test_girder_worker |   File "/girder_worker/girder_worker/plugins/girder_io/__init__.py", line 1, in <module>
gw_integration_test_girder_worker |     import girder_client
gw_integration_test_girder_worker | ImportError: No module named girder_client
gw_integration_test_girder_worker exited with code 1
```

cjh1:
@kotfic Is this ready for another look?

cjh1:
@kotfic Still seems to be using default ports for mongo and rabbbit? Can we switch to something else, so we can run without have to stop them of the host machine?

tests/integration/Dockerfile.girder_worker
```diff
@@ -0,0 +1,19 @@
+FROM girder/girder_worker:latest
+MAINTAINER Christopher Kotfila <chris.kotfila@kitware.com>
+
+USER root
+
+RUN sed -i  's|^broker.*=.*$|broker = amqp://guest:guest@rabbit/|' /girder_worker/girder_worker/worker.local.cfg
+
+RUN apt update && apt install -y sudo
```
danlamanna: apt shouldn't be run from a CLI, use apt-get instead.

tests/integration/Dockerfile.girder
```diff
@@ -0,0 +1,29 @@
...
+# e.g. for developing test endpoints
+VOLUME /girder/girder/plugins/integration_test_endpoints
+
+# Allow moutning /girder_worker
+RUN mkdir /girder_worker
+VOLUME /girder_worker
+
+# Make sure remote debugging is available
+RUN pip install rpdb
+
```
danlamanna: Should the logs be mounted somehow? I can imagine after the tests we would want to place Girder error/info logs into Circle artifacts.

tests/integration/conftest.py
```diff
@@ -0,0 +1,113 @@
...
+from girder_worker.utils import JobStatus
+from datetime import timedelta, datetime
+
+
+def pytest_addoption(parser):
+        parser.addoption('--girder', action='store', default='http://127.0.0.1:8989',
+                         help='my option: type1 or type2')
+
+
+@pytest.fixture(scope='module')
```
danlamanna: fixtures should be in their own file(s) separate from configuration.

tests/integration/conftest.py
```diff
@@ -0,0 +1,113 @@
...
+    return functools.partial(
+        wait_for,
+        predicate=lambda j: j['status'] == JobStatus.ERROR,
+        on_timeout=on_timeout)
+
+
+# pytest hook for ordering test items after they have been
+# collected. This uses 'priority' mark objects to order the
+# tests, defaulting the priority of non-marked tests to 100.
+def pytest_collection_modifyitems(items):
```
danlamanna: Look into pytest-dependency for something like this.

cjh1:
@kotfic After following the instructions I get:

```
  File "/home/cjh/virtualenv/girder_worker_tests/local/lib/python2.7/site-packages/_pytest/config.py", line 367, in _importconftest
    mod = conftestpath.pyimport()
  File "/home/cjh/virtualenv/girder_worker_tests/local/lib/python2.7/site-packages/py/_path/local.py", line 662, in pyimport
    __import__(modname)
  File "/home/cjh/virtualenv/girder_worker_tests/local/lib/python2.7/site-packages/_pytest/assertion/rewrite.py", line 216, in load_module
    py.builtin.exec_(co, mod.__dict__)
  File "/home/cjh/virtualenv/girder_worker_tests/local/lib/python2.7/site-packages/py/_builtin.py", line 221, in exec_
    exec2(obj, globals, locals)
  File "<string>", line 7, in exec2
  File "/home/cjh/work/source/girder_worker/tests/integration/conftest.py", line 7, in <module>
    from girder_worker.utils import JobStatus
ImportError: No module named girder_worker.utils
ERROR: could not load /home/cjh/work/source/girder_worker/tests/integration/conftest.py
```
Looks like girder_worker needs to be install in host environment as well?

kotfic:
I have recently pushed docker container images to dockerhub.  The new way to test this is as follows:

Create a virtual environment and cd into a working directory 
```sh
# Install girder-worker
git clone https://github.com/girder/girder_worker.git .
git checkout integration-test-framework
pip install -e .

# Run the integration environment
cd tests/integration
docker-compose up

# From a separate terminal - Run the test suite
cd tests/integration
pip install -r requirements.txt
pytest -v  
```

tests/integration/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: '2'
+services:
+  girder:
+    image: girder/girder:gw_integration_test
```
danlamanna: Can build be used here as well? So `docker-compose build` can be used to build all the relevant Dockerfiles instead of pulling?

tests/integration/Dockerfile.girder
```diff
@@ -0,0 +1,29 @@
+FROM girder/girder:latest
+MAINTAINER Christopher Kotfila <chris.kotfila@kitware.com>
+
+VOLUME /girder/plugins/integration_test_endpoints
+
+RUN sed -i 's|^uri = "mongodb://localhost:27017/girder"$|uri = "mongodb://mongo:27017/girder"|' /girder/girder/conf/girder.local.cfg
+RUN sed -i 's|^server.socket_port = 8080$|server.socket_port = 8989|' /girder/girder/conf/girder.local.cfg
```
danlamanna: Should these just be moved to the entrypoint since they're both flags `python -m girder` takes? I think that's less brittle than sed.

tests/integration/Dockerfile.girder_worker
```diff
@@ -0,0 +1,19 @@
+FROM girder/girder_worker:latest
+MAINTAINER Christopher Kotfila <chris.kotfila@kitware.com>
+
+USER root
+
+RUN sed -i  's|^broker.*=.*$|broker = amqp://guest:guest@rabbit/|' /girder_worker/girder_worker/worker.local.cfg
```
danlamanna: I think the same comment from above applies here as well.

tests/integration/Dockerfile.girder
```diff
@@ -0,0 +1,29 @@
...
+# Allow mounting integration_test_endpoints from host environment
+# e.g. for developing test endpoints
+VOLUME /girder/girder/plugins/integration_test_endpoints
+
+# Allow moutning /girder_worker
+RUN mkdir /girder_worker
+VOLUME /girder_worker
+
+# Make sure remote debugging is available
+RUN pip install rpdb
```
danlamanna: Are these `rpdb` installs staying in? I think it's fine to leave them in just want to make sure.

tests/integration/common_tasks/setup.py
```diff
@@ -0,0 +1,24 @@
+from setuptools import setup
+
+setup(name='common_tasks',
+      version='0.0.0',
+      description='An example girder worker extension',
+      author='Chris Kotfila',
```
danlamanna: Kitware

tests/integration/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def pytest_addoption(parser):
+    parser.addoption('--girder', action='store', default='http://127.0.0.1:8989/',
+                     help='Specify a different server to run against')
+
+
+@pytest.fixture(scope='module')
+def api_url(request):
+    return request.config.getoption('--girder') + 'api/v1/'
```
danlamanna: I don't think `api/v1` should be hardcoded here.

tests/integration/common_tasks/common_tasks/test_tasks/fib.py
```diff
@@ -0,0 +1,19 @@
...
+    return fibonacci(n-1) + fibonacci(n-2)
+
+
+@app.task(bind=True)
+def fib_seq(task, n, **kwargs):
+    if n < 0:
+        raise Exception('Must pass in positive integer!')
+
+    for _n in range(1, n+1):
+        print '%s: %s' % (_n, fibonacci(_n))
```
danlamanna: We should make sure these tests run under Python 3 before we merge, to avoid adding more work to #119.

tests/integration/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+# TODO: test with admin and non-admin user - are there (should there be)
+# differences between girder-worker functionality between the two?
+@pytest.fixture(scope='module',
+                params=[['admin', 'letmein']],
+                ids=['admin'])
+def session(request, api_url):
+    username, password = request.param
+    with GirderSession(base_url=api_url) as s:
+        try:
+            r = s.get('user/authentication',
```
danlamanna: I think this can be on one line.

tests/integration/integration_test_endpoints/server/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+        self.girder_worker_run_failing_analysis['script'] = 'this should fail'
+
+        self.girder_worker_run_inputs = {'a': {'format': 'integer', 'data': 1},
+                                         'b': {'format': 'integer', 'data': 2}}
+
+        self.girder_worker_run_outputs = {'c': {'format': 'integer'}}
+
+    @access.token
+    @describeRoute(
+        Description('Utility endpoint to get an asyn result from a celery id')
```
danlamanna: async

tests/integration/Dockerfile.girder
```diff
@@ -0,0 +1,29 @@
...
+
+
+COPY ./integration_test_endpoints /girder/girder/plugins/
+RUN girder-install web --plugins=integration_test_endpoints
+
+# Allow mounting integration_test_endpoints from host environment
+# e.g. for developing test endpoints
+VOLUME /girder/girder/plugins/integration_test_endpoints
+
+# Allow moutning /girder_worker
```
danlamanna: mounting

tests/integration/common_tasks/common_tasks/test_tasks/fib.py
```diff
@@ -0,0 +1,19 @@
...
+@girder_job(title='Fibonacci Job')
+@app.task
+def fibonacci(n, **kwargs):
+    if n == 1 or n == 2:
+        return 1
+    return fibonacci(n-1) + fibonacci(n-2)
+
+
+@app.task(bind=True)
+def fib_seq(task, n, **kwargs):
```
danlamanna: This is a silly point, but I've noticed we use `task` when all of the Celery documentation I've seen uses `self` when defining bound tasks, should we start transitioning towards that moving forward?

tests/integration/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest
```
danlamanna: Does it make sense to use `tests_requires` in `setup.py` for this?

tests/integration/utilities.py
```diff
@@ -0,0 +1,56 @@
...
+            on_timeout=lambda j:
+                'Timed out waiting for job/%s to move into error state' % j['_id'])
+
+    def get_result(self, celery_id):
+        r = self.post('integration_tests/result', data={
+            'celery_id': celery_id})
+        return r.text
+
+    @contextlib.contextmanager
+    def wait_for(self, job_id, predicate, timeout=20, interval=0.3, on_timeout=None):
```
danlamanna: Docstring

tests/integration/utilities.py
```diff
@@ -0,0 +1,56 @@
...
+        then = datetime.utcnow() + timedelta(seconds=timeout)
+        timeout = True
+
+        while datetime.utcnow() < then:
+            r = self.get('job/' + job_id)
+
+            if predicate(r.json()):
```
danlamanna: How should the case where `not r.ok` be handled?

tests/integration/Dockerfile.girder
kotfic: no problem

tests/integration/Dockerfile.girder
kotfic: I was going to leave it in,  i end up needing it plenty of times,  and I think other people will too.

tests/integration/Dockerfile.girder_worker
kotfic: unfortunately while you can pass ```-b``` on the command line to the girder_worker process,  when you import a task in the girder context to use ```.delay()``` or whatever,  it won't have this  broker,  so it has to be changed in the config file anyways

danlamanna:
When running this in parallel with xdist, I was somewhat frequently able to reproduce a failure that (so far) seems to be resolved by https://github.com/girder/girder/pull/2110.

PR #151, opened 2017-06-05
Use raise_for_status() to check status code

cjh1:
We need to check that our requests back to Girder are successful!

Found this when trying to debug cancellation functionality ....

### DIFF
--- a/girder_worker/utils.py
+++ b/girder_worker/utils.py
@@ -166,7 +166,7 @@ class JobManager(object):
         self.status = status
         self._redirectPipes(False)
         req = requests.request(self.method.upper(), self.url, headers=self.headers,
-                         data={'status': status}, allow_redirects=True)
+                               data={'status': status}, allow_redirects=True)
         req.raise_for_status()
         self._redirectPipes(True)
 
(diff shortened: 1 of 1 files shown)


girder_worker/utils.py
```diff
@@ -107,14 +107,15 @@ def _flush(self):
...
-            requests.request(
+            req = requests.request(
                 self.method.upper(), self.url, allow_redirects=True,
                 headers=self.headers, data={
                     'log': self._buf,
                     'progressTotal': self._progressTotal,
                     'progressCurrent': self._progressCurrent,
                     'progressMessage': self._progressMessage
                 })
+            req.raise_for_status()
```
zachmullen: If _flush throws an exception, that doesn't put us into some kind of infinite recursion, does it?

cjh1: I don't think so, I will fake an exception and see what happens, think the exception would just bubble up to celery.

zachmullen: I think you're right, just wanted to sanity check, since this gets called during exception handling.

cjh1: Experimentation shows it just gets caught by Celery, we are good I think.

PR #153, opened 2017-06-07
WIP: Support nvidia-docker for running hardware-accelerated containers

zachmullen:


zachmullen:
@girder/developers I'd like to get some consensus on client controls over nvidia docker usage. As written, this code introspects the container, detects if it requires nvidia-docker support, and if so, configures the container to run with the nvidia bindings. This means any nvidia-based container will require the worker node to be running the nvidia-docker plugin in its docker engine. The question is, should we have some sort of control in the task spec to explicitly enable or disable the nvidia behavior, and what should be its default value?

We could also defer this feature until later if we don't think it's important in the first pass.

cjh1:
@zachmullen What happens if you try to run with the nvidia bindings  without the nvidia-docker plugin installed in the docker engine? I would say that if the container required nvidia-docker support and the engine has this support( can we detect this a run-time? ) we should just do the right think. Is there any reason as user would want to disable this behavior?

zachmullen:
Agreed generally.

> Is there any reason as user would want to disable this behavior?

The only use case I can think of is a user who has a nvidia-based image, but who wants to run something within it that does not require nvidia drivers. For that case, I was thinking the currently written behavior would be the default, and an option of `disable_nvidia_support=True` could be passed to override it.

cjh1:
That sounds like a good plan, but I think the current behavior is fine for the moment.

girder_worker/plugins/docker/executor.py
```diff
@@ -163,7 +164,7 @@ def make_pipe(id, spec, bindings):
 
 def _run_container(image, args, **kwargs):
     # TODO we could allow configuration of non default socket
-    client = docker.from_env(version='auto')
+    client = nvidia.get_client_for_image(image)
```
mgrauer: This is surprising to read.  We are in _run_container, then we call nvidia.get_client_for_image whether I'm using an nvidia Docker or not.  Maybe a comment here or a different method name than get_client_for_image would make this more clear, that I'm only getting nvidia client when appropriate.  

or maybe something like 

```
client = DockerClient.from_env(version='auto')
if nvidia.is_nvidia_image(client.api, image):
  client = nvidia.get_nvidia_client()
```

zachmullen:
@mgrauer you are right, fixed.

mgrauer:
Needs tests.

zachmullen:
How do we feel about merging this without tests? I've been promising that this support exists to multiple groups now, but it's not actually in master...

mgrauer:
I'm ok with it going it, as long as it doesn't break any of the standard paths, and those paths are actually tested, even if this one isn't tested.

Is this fully supported, or do you want to mark it preliminary somehow?  It doesn't matter from our versioning perspective, because we are major version 0 still, but I think it would be good to add some remarks to the top of `nvidia.py` describing what our intention and level of commitment is to the approach.  Maybe we can even build in a decision point for later, like we won't add it to a 1.0 release unless tested, or will add it as an explicitly untested and therefore flaky feature.

zachmullen:
AFAIC, this is fully supported. It can only truly be integration tested on a machine with a GPU, which rules out CI. Mock testing would be doable, but a lot of work since I could not get dockerpy to play nice with vcr or betamax.

PR #156, opened 2017-06-12
Add $output{} template expansion for dynamic output filenames

zachmullen:
Still need to add testing

### DIFF
--- a/girder_worker/plugins/docker/executor.py
+++ b/girder_worker/plugins/docker/executor.py
@@ -71,6 +71,8 @@ def _expand_args(args, inputs, taskInputs, outputs, tmpDir):
                 arg = arg.replace('$input{%s}' % inputId, str(transformed))
             elif inputId == '_tempdir':
                 arg = arg.replace('$input{_tempdir}', DATA_VOLUME)
+            else:
+                raise Exception('Could not expand token: $input{%s}.' % inputId)
         for inputId in _flagRe.findall(arg):
             if inputId in inputs and inputs[inputId]['script_data']:
                 val = taskInputs[inputId].get('arg', inputId)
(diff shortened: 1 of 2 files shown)


girder_worker/plugins/docker/executor.py
```diff
@@ -43,35 +47,44 @@ def _transform_path(inputs, taskInputs, inputId, tmpDir):
...
     for arg in args:
         skip = False
-        for inputId in re.findall(inputRe, arg):
+        for inputId in _inputRe.findall(arg):
             if inputId in inputs:
                 transformed = _transform_path(
                     inputs, taskInputs, inputId, tmpDir)
                 arg = arg.replace('$input{%s}' % inputId, str(transformed))
             elif inputId == '_tempdir':
                 arg = arg.replace('$input{_tempdir}', DATA_VOLUME)
```
jeffbaumes: Should there be an `else` here in parallel to the new output loop below that raises an exception?

zachmullen: The parallel exception is sort of handled in `_transform_path` for the case of inputs.

zachmullen: Though, I guess that exception wouldn't be reached since we are inside `if inputId in inputs`.  Which behavior is more desirable in this: raising an exception, or just leaving the template value unexpanded? I'll make it the same for both.

jeffbaumes: Ah, ok. It seems there is a bit more nuance to `$input{}` so that the code paths cannot parallel each other or use the same code? I see that e.g. `os.path.join(DATA_VOLUME, ...)` is used in both places.

zachmullen:
This is ready for review.

zachmullen:
@jeffbaumes the last commit addresses the two changes we discussed, PTAL.

jeffbaumes:
@zachmullen merge when ready.

PR #157, opened 2017-06-14
Allow to configure celery's backend

Xarthisius:
The recent version of celery is complaining when `backend` is set to amqp and wants to use `rpc://` instead. It would be nice if `backend` would be configurable.

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -7,6 +7,7 @@ from celery.signals import (task_prerun, task_postrun,
                             task_failure, task_success,
                             worker_ready, before_task_publish)
 from celery.result import AsyncResult
+from six.moves import configparser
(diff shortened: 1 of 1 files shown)


girder_worker/app.py
```diff
@@ -262,7 +262,7 @@ class _CeleryConfig:
 
 app = Celery(
     main=girder_worker.config.get('celery', 'app_main'),
-    backend=girder_worker.config.get('celery', 'broker'),
+    backend=girder_worker.config.get('celery', 'backend'),
```
jeffbaumes: For backwards compatibility, should backend default to the broker setting if backend is not provided?

zachmullen:
Thanks @Xarthisius !

PR #162, opened 2017-06-26
Add parameter description decorators compatible with item_tasks

### LINKED ISSUE #143: Implement self-describing Celery task API
### Deliverables:
+ tasks imported into girder should have a property, or function,  that produces an item_task comparable 'spec'
### Technical Notes:
+ Provide input/output spec metadata on a decorator for celery task
  + Statically defined data will be stored in custom subclass of celery.app.Task (e.g. girder_worker.app.Task)
  + For docker_run task we can use a custom subclass of the girder_worker Task class (e.g. girder_worker.app.DockerTask)

jbeezley:
Fixes #143 

From the canonical fibonacci example:
```python
from girder_worker.app import app


@app.task
@app.argument('n', app.types.Integer, min=1)
def fibonacci(n):
    """Compute the nth fibonacci number recursively."""
    if n == 1 or n == 2:
        return 1
    return fibonacci(n-1) + fibonacci(n-2)
```

Then
```python
>>> fibonacci.describe()
{'description': 'Compute the nth fibonacci number recursively.',
 'inputs': [{'description': 'Select a number',
   'id': 'n',
   'min': 1,
   'name': 'n',
   'step': 1,
   'type': 'integer'}],
 'name': u'tasks.fibonacci'}
```

Also, it can run from `item_tasks` input bindings
```python
>>> fibonacci.call_item_task({'n': {'data': 5}})
5
```

There are a couple of things missing that I can add here or in a new PR.  Before I get too deep into it, it would be good to get feedback on this implementation.
1. Handle file inputs
2. Handle outputs


jeffbaumes:
I like where this is going. One thing to note is that there may be some confusion of the new `types` dir and `plugins\types`.

Another thing is whether we can make the decorator more like the `Description` decorator in Girder https://github.com/girder/girder/blob/master/girder/api/describe.py to keep a corresponding mental model on how to annotate python function parameter types when moving between Girder and Girder Worker.

We also discussed making the decorators also work to support JSON and command-line generation outside a celery context, to help make Docker-but-not-Celery based item-tasks, e.g. `pysciencedock`.

jbeezley:
> One thing to note is that there may be some confusion of the new types dir and plugins\types.

I can see that, any suggestions on how to rename it?

> Another thing is whether we can make the decorator more like the Description decorator in Girder https://github.com/girder/girder/blob/master/girder/api/describe.py to keep a corresponding mental model on how to annotate python function parameter types when moving between Girder and Girder Worker.

You're right, it is a departure from girder's decorators.  I personally prefer this style over what girder does.  It looks a lot more like click and celery decorators, but I change it if that's what others prefer.

> We also discussed making the decorators also work to support JSON and command-line generation outside a celery context, to help make Docker-but-not-Celery based item-tasks, e.g. pysciencedock.

Good point.  It will be a pretty quick change to remove the dependence on celery tasks and allow this to work bare functions.

jeffbaumes:
Is it too wild to implement this in a way such that the decorated function has a `.json()` that returns an item_tasks descriptor, a `.cli()` that returns a (click-based?) arg parser, and a `.task()` that returns a Celery task object? Function names are up for discussion.

kotfic:
Looks like we need to add funcsigs to the required packages?  With python 2.7.13 I get the following:

```
In [3]: from gwexample.analyses.tasks import fibonacci
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-3-9f1ef39da8de> in <module>()
----> 1 from gwexample.analyses.tasks import fibonacci

/home/kotfic/kitware/review/girder_worker/item_tasks_description/examples/plugin_example/gwexample/analyses/tasks.py in <module>()
----> 1 from girder_worker.app import app
      2 
      3 
      4 @app.task
      5 @app.argument('n', app.types.Integer, min=1)

/home/kotfic/kitware/review/girder_worker/item_tasks_description/girder_worker/app.py in <module>()
     11 import sys
     12 from .utils import JobStatus
---> 13 from describe import argument, describe_task, parse_inputs
     14 import types
     15 

/home/kotfic/kitware/review/girder_worker/item_tasks_description/girder_worker/describe.py in <module>()
      5     from inspect import signature
      6 except ImportError:
----> 7     from funcsigs import signature
      8 
      9 

ImportError: No module named funcsigs
```

jbeezley:
Right, I forgot to add that.

girder_worker/describe.py
```diff
@@ -0,0 +1,113 @@
...
+
+def get_description_attribute(func):
+    """Get the private description attribute from a function."""
+    description = getattr(func, '_girder_description', None)
+    if description is None:
+        raise MissingDescriptionException('Task is missing description decorators')
+    return description
+
+
+def argument(name, type, *args, **kwargs):
```
kotfic: Sorry to be nit picky - but can we avoid using reserved words as argument names?

girder_worker/describe.py
```diff
@@ -0,0 +1,113 @@
...
+        'inputs': inputs
+    }
+    desc = description.get('description', getdoc(func))
+    if desc:
+        spec['description'] = desc
+
+    return spec
+
+
+def get_input_data(arg, input):
```
kotfic: avoid using ```input``` reserved word

girder_worker/describe.py
```diff
@@ -0,0 +1,113 @@
...
+    :returns: args and kwargs objects to call the task with
+    """
+    func = task.run
+    description = get_description_attribute(func)
+    arguments = description.get('arguments', [])
+    args = []
+    kwargs = {}
+    for arg in arguments:
+        desc = arg.describe()
+        id = desc['id']
```
kotfic: avoid using ```id``` reserved word

girder_worker/types/base.py
```diff
@@ -0,0 +1,67 @@
...
+        copy.setdefault('name', self.name)
+        if self.has_default():
+            copy.setdefault('default', {
+                'data': self.serialize(self.parameter.default)
+            })
+        if self.help is not None:
+            copy['description'] = self.help
+        return copy
+
+    def serialize(self, value):
```
kotfic: Sorry,  i'm just a little confused about these serialize/deserialize functions?  I would have expected serialize to convert the object into a JSON-afiable dict (and take only ```self```) as argument,  and for deserialize to essentially be a static method that takes that same dict and returns the object.  Could you help me understand?

girder_worker/app.py
```diff
@@ -246,4 +255,6 @@ class _CeleryConfig:
     backend=backend_uri, broker=broker_uri,
     task_cls='girder_worker.app:Task')
 
+app.argument = argument
```
kotfic: Let's keep the argument decorator (and the rest of the 'describe' api) completely separate from the celery object and just directly assign _girder_description to the function objects.  that way this can be generically used across different applications (e.g. celery,  docker,  cli,  etc).

girder_worker/describe.py
manthey: To be nit-picky, `type`, `input`, and `id` are NOT reserved words in python.  They are just references to built-in functions that can be redefined in any scope one likes.  Things like `and` and `as` are reserved words.  See Python [2](https://docs.python.org/2/reference/lexical_analysis.html#keywords) and [3](https://docs.python.org/3/reference/lexical_analysis.html#keywords) for a list of actual reserved words.

fogleman: "Avoid shadowing builtins."?

girder_worker/types/base.py
jbeezley: Yes, they aren't fully fleshed out.  They are intended to serialize a **value** of the given type rather than the type descriptor class.  The motivation here is that item_tasks (by virtue of it's origins from the slicer execution module) passes parameters in a way that isn't very pythonic.  For example, "vector" types, rather than being a list, are concatenated into a comma separated string.  I can see other uses in terms of normalization of values before being sent to the client (colors into an rgb string for example) and handling file types.

jbeezley:
I made an initial attempt at adding input parameter support for items and folders in https://github.com/girder/girder_worker/pull/162/commits/65eebb6a84aeb265f0d18c590ae763ae104f1aef.  The UI widget provided to the user will be the standard browser widget.  The value passed to the function will be a Girder document id as a string.

This is intended to handle the generic use case where it will be the function implementer's responsibility to do something useful with the id.  I think this assumes that the function is wrapped in a celery task with an instantiated girder client.  I'm not sure how useful it is outside of that context.  I'm open to suggestions for how better to handle girder objects.

Issue #161, opened 2017-06-23
Full documentation overhaul


mgrauer:
There's no reason this couldn't start before the full refactor, at least some paragraphs outlining the vision of Girder-Worker and what are the major paths.

mgrauer:
Let's be sure to remove any data conversion and validation stuff that is going away, along with any modes that have been dropped.  Let's also update the copyright from 2014 in the read the docs.  At least to 2015.

PR #164, opened 2017-07-06
Refactor custom job options

kotfic:
This provides a cleaner approach to passing custom girder job options
through the girder_worker.app.Task object's implementation of
apply_async so that before_task_publish has access to special
arguments used for job creation,  setting the girder_api_url and
girder_client_token.

This introduces two class level variables special_headers and
special_options. special_options will not be included in the headers
sent on the message, special_headers will. In both cases if these keys
appear in delay()'s kwargs, or in apply_async()'s options they will be
popped out and placed in headers so before_task_publish can use them
to generate a girder job.

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -47,7 +47,7 @@ class Task(celery.Task):
     """Girder Worker Task object"""
 
     _girder_job_title = '<unnamed job>'
-    _girder_job_type = None
+    _girder_job_type = 'celery'
     _girder_job_public = False
     _girder_job_handler = 'celery_handler'
     _girder_job_other_fields = {}
@@ -77,7 +77,15 @@ class Task(celery.Task):
 
         # Pass girder related job information through to
         # the signals by adding this information to options['headers']
-        headers = {}
+        # This sets defaults for reserved_options based on the class defaults,
+        # or values defined by the girder_job() dectorator
+        headers = {
+            'girder_job_title': self._girder_job_title,
+            'girder_job_type': self._girder_job_type,
+            'girder_job_public': self._girder_job_public,
+            'girder_job_handler': self._girder_job_handler,
+            'girder_job_other_fields': self._girder_job_other_fields,
+        }
 
         # Certain keys may show up in either kwargs (e.g. via .delay(girder_token='foo')
         # or in options (e.g.  .apply_async(args=(), kwargs={}, girder_token='foo')
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -46,13 +46,27 @@ class GirderAsyncResult(AsyncResult):
 class Task(celery.Task):
     """Girder Worker Task object"""
 
-    _girder_job_title = None
+    _girder_job_title = '<unnamed job>'
     _girder_job_type = None
     _girder_job_public = False
     _girder_job_handler = 'celery_handler'
     _girder_job_other_fields = {}
 
-    special_headers = ['girder_token', 'girder_user']
+    # These keys will be removed from apply_async's kwargs or options and
+    # transfered into the headers of the message.
+    special_headers = [
+        'girder_client_token',
+        'girder_api_url']
+
+    # These keys will be available in the 'properties' dictionary inside
+    # girder_before_task_publish() but will not be passed along in the message
+    special_options = [
+        'girder_user',
+        'girder_job_title',
+        'girder_job_type',
+        'girder_job_public',
+        'girder_job_handler',
+        'girder_job_other_fields']
 
     def AsyncResult(self, task_id, **kwargs):
         return GirderAsyncResult(task_id, backend=self.backend,
@@ -69,17 +83,11 @@ class Task(celery.Task):
         # or in options (e.g.  .apply_async(args=(), kwargs={}, girder_token='foo')
         # For those special headers,  pop them out of kwargs or options and put them
         # in headers so they can be picked up by the before_task_publish signal.
-        for key in self.special_headers:
+        for key in self.special_headers + self.special_options:
             if kwargs is not None and key in kwargs:
                 headers[key] = kwargs.pop(key)
             if key in options:
                 headers[key] = options.pop(key)
-
-        headers['girder_job_title'] = self._girder_job_title
-        headers['girder_job_type'] = self._girder_job_type
-        headers['girder_job_public'] = self._girder_job_public
-        headers['girder_job_handler'] = self._girder_job_handler
-        headers['girder_job_other_fields'] = self._girder_job_other_fields
 
         if 'headers' in options:
             options['headers'].update(headers)
@@ -118,34 +126,53 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
             job_model = ModelImporter.model('job', 'jobs')
 
             user = headers.pop('girder_user', getCurrentUser())
-            token = headers.pop('girder_token', None)
-
             task_args, task_kwargs = body[0], body[1]
 
             job = job_model.createJob(
-                **{'title': headers.get('girder_job_title', Task._girder_job_title),
-                   'type': headers.get('girder_job_type', Task._girder_job_type),
-                   'handler': headers.get('girder_job_handler', Task._girder_job_handler),
-                   'public': headers.get('girder_job_public', Task._girder_job_public),
+                **{'title': headers.pop('girder_job_title', Task._girder_job_title),
+                   'type': headers.pop('girder_job_type', Task._girder_job_type),
+                   'handler': headers.pop('girder_job_handler', Task._girder_job_handler),
+                   'public': headers.pop('girder_job_public', Task._girder_job_public),
                    'user': user,
                    'args': task_args,
                    'kwargs': task_kwargs,
                    'otherFields': dict(celeryTaskId=headers['id'],
-                                       **headers.get('girder_job_other_fields',
+                                       **headers.pop('girder_job_other_fields',
                                                      Task._girder_job_other_fields))})
-            # If we don't have a token from girder_token kwarg,  use
-            # the job token instead. Otherwise no token
-            if token is None:
-                token = job.get('token', None)
 
-            headers['jobInfoSpec'] = utils.jobInfoSpec(job, token)
-            headers['apiUrl'] = utils.getWorkerApiUrl()
+            headers['jobInfoSpec'] = utils.jobInfoSpec(job)
 
         except ImportError:
             # TODO: Check for self.job_manager to see if we have
             #       tokens etc to contact girder and create a job model
             #       we may be in a chain or a chord or some-such
             pass
+
+    if 'girder_api_url' not in headers:
+        try:
+            from girder.plugins.worker import utils
+            headers['girder_api_url'] = utils.getWorkerApiUrl()
+        except ImportError:
+            # TODO: handle situation where girder_worker is producing the message
+            #       Note - this may not come up at all depending on how we pass
+            #       girder_api_url through to the next task (e.g. in the context
+            #       of chaining events)
+            pass
+
+    if 'girder_client_token' not in headers:
+        try:
+            from girder.utility.model_importer import ModelImporter
+            headers['girder_client_token'] = ModelImporter.model('token').createToken()
+        except ImportError:
+            # TODO: handle situation where girder_worker is producing the message
+            #       Note - this may not come up at all depending on how we pass
+            #       girder_token through to the next task (e.g. in the context
+            #       of chaining events)
+            pass
+
+    # Finally,  remove all special_options from headers
+    for key in Task.special_options:
+        headers.pop(key)
 
 
 @worker_ready.connect
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -10,6 +10,7 @@ from celery.result import AsyncResult
 from six.moves import configparser
 import sys
 from .utils import JobStatus
+from girder_client import GirderClient
 
 
 class GirderAsyncResult(AsyncResult):
@@ -234,6 +235,12 @@ def gw_task_prerun(task=None, sender=None, task_id=None,
     except JobSpecNotFound:
         task.job_manager = None
         print('Warning: No jobInfoSpec. Setting job_manager to None.')
+
+    try:
+        task.girder_client = GirderClient(apiUrl=task.request.girder_api_url)
+        task.girder_client.token = task.request.girder_client_token
+    except AttributeError:
+        task.girder_client = None
 
 
 @task_success.connect
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -54,13 +54,13 @@ class Task(celery.Task):
 
     # These keys will be removed from apply_async's kwargs or options and
     # transfered into the headers of the message.
-    special_headers = [
+    reserved_headers = [
         'girder_client_token',
         'girder_api_url']
 
     # These keys will be available in the 'properties' dictionary inside
     # girder_before_task_publish() but will not be passed along in the message
-    special_options = [
+    reserved_options = [
         'girder_user',
         'girder_job_title',
         'girder_job_type',
@@ -83,7 +83,7 @@ class Task(celery.Task):
         # or in options (e.g.  .apply_async(args=(), kwargs={}, girder_token='foo')
         # For those special headers,  pop them out of kwargs or options and put them
         # in headers so they can be picked up by the before_task_publish signal.
-        for key in self.special_headers + self.special_options:
+        for key in self.reserved_headers + self.reserved_options:
             if kwargs is not None and key in kwargs:
                 headers[key] = kwargs.pop(key)
             if key in options:
@@ -170,9 +170,9 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
             #       of chaining events)
             pass
 
-    # Finally,  remove all special_options from headers
-    for key in Task.special_options:
-        headers.pop(key)
+    # Finally,  remove all reserved_options from headers
+    for key in Task.reserved_options:
+        headers.pop(key, None)
 
 
 @worker_ready.connect
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ pytz==2016.4
 requests[security]==2.10.0
 six==1.10.0
 stevedore==1.24.0
+girder-client==2.2.1
--- a/tests/integration/HACKING.md
+++ b/tests/integration/HACKING.md
@@ -1,12 +1,16 @@
+*Note: This documentation is a work in progress, if you find problems with it's construction or it's content.  Please file an issue!*
+
 # What is this all about?
-In this documentation we will descend through increasingly detailed descriptions of Girder Worker and it's relationship to [Celery](http://www.celeryproject.org/), ending up with descriptions of how you can use Girder Worker to run remote, asynchronous, tasks in your own extensions to Girder.
+In this documentation we will descend through increasingly detailed descriptions of Girder Worker and its relationship to [Celery](http://www.celeryproject.org/), ending up with descriptions of how you can use Girder Worker to run remote asynchronous tasks.
 
 ## 10,000 foot view
 Girder Worker is an application that wraps Celery; it relies on Celery's distributed task execution mechanisms to do work, while reflecting task state and logging information back into [Girder](https://girder.readthedocs.io/) for visualization, debugging and management. Girder Worker is designed to receive and execute tasks that have been produced by Girder. Messages that encode what tasks and task arguments are passed from Girder to Girder Worker via a separate service called a [broker](http://docs.celeryproject.org/en/latest/getting-started/brokers/). In many common setups this broker is [RabbitMQ](https://www.rabbitmq.com/). 
 
 Girder Worker runs in a separate process from Girder. It may run on the same physical machine as Girder,  or may be run on one or more separate machines. At a low level, RabbitMQ brokers the communication between Girder and Girder Worker.  If Girder and Girder Worker are running on separate machines, both machines must be able to connect to the server running RabbitMQ.
 
-Assuming you have configured girder-worker to connect to the RabbitMQ service, running Girder Worker is easy. From a command line simply run:
+Girder Worker's configuration is managed by an INI style configuration file located in ```girder_worker/worker.dist.cfg```. This file is version controlled and should not be directly modified.  To customize configurations copy the ```worker.dist.cfg``` to ```worker.local.cfg``` and make your modifications there. There are several commented configurations available in the configuration file. The most important is the ```broker``` config under the ```[celery]``` section. This is how Girder Worker connects to RabbitMQ and should follow Celery's [broker_url](http://docs.celeryproject.org/en/latest/userguide/configuration.html#broker-url) syntax. 
+
+Assuming you have configured Girder Worker to connect to the RabbitMQ service, running Girder Worker is easy. From a command line simply run:
 
 ```python
 $> girder-worker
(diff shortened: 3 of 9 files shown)


mgrauer:
Provide examples of how to use these new paths in readthedocs as part of #161, can be done post-merge.

mgrauer:
Add a temporary readme that can be a basis for doc overhaul later.

jeffbaumes:
I mildly prefer something more generic sounding like "extra" or "custom" instead of "special" for the field names.

cjh1:
how about "bespoke" :smile:

kotfic:
how about 'reserved'?  It is worth noting that these will conflict with functions that try to define their own ```girder_job_title``` keyword arguments

girder_worker/app.py
```diff
@@ -109,35 +117,54 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+        except ImportError:
+            # TODO: handle situation where girder_worker is producing the message
+            #       Note - this may not come up at all depending on how we pass
+            #       girder_token through to the next task (e.g. in the context
+            #       of chaining events)
+            pass
+
+    # Finally,  remove all special_options from headers
+    for key in Task.special_options:
+        headers.pop(key)
```
danlamanna: Should we actually remove these? I can't think of a use case off the top of my head, but could having these available as headers on the message be useful inside the task?

kotfic:
This supersedes https://github.com/girder/girder_worker/pull/104

girder_worker/app.py
kotfic: I could go either way on it really - i've leaned towards no because i think its better to have the job model in girder be the oracle of this data (rather than reference the headers of the amqp message inside the task).  The nice thing is,  if there is a compelling reason to keep something in the headers it just needs to be moved from ```reserved_options```  to ```reserved_headers``` inside the ```girder_worker.app.Task``` class

kotfic:
@mgrauer PTAL at the [small book](https://github.com/girder/girder_worker/blob/header-options-support/tests/integration/HACKING.md) i just wrote. This is not really where most of this documentation belongs,  but as we figure #161 out I thought I would be good to just do a brain dump. Thanks!

zachmullen:
Nice docs @kotfic -- just a reminder that we are in need of a major overhaul of our RTD docs, so if any of this content should be part of that, we may want to consider making these changes part of that larger picture.

kotfic:
@zachmullen yes definitely - I see this as being purely transitional. I'm hoping this stuff will end up being apart of the overhaul,  but until we figure out exactly what that is going to look like,  i don't want refactor details to be based on tribal knowledge.

jeffbaumes:
@kotfic thanks for the thoughtful explanation. From your book:

> The fundamental issue with the Traditional Method is how does one synchronize Celery's configuration across producer (Girder) and consumer (Girder Worker) processes, keeping in mind that these processes may be taking place on two (or more) different physical machines. Traditionally Celery solves this problem through the Python import mechanism.

This helps answer a question that has been nagging me: Requiring importing the task package has some clear downsides, but what are the upsides?

Do you envision supporting the "traditional approach" long term or deprecating it? I see the approaches as two sides of a coin. The traditional approach starts with a Girder Job and crams in all Celery Task information in a special, non-standard way. The new approach starts with a Celery Task and crams in all Girder Job information in a special, slightly non-standard way. The (good) argument is that Celery Tasks are more complex and nuanced than Girder Jobs, so we should normally be starting with the Celery Task.

I think some of the exposition you provide about how Girder and Girder Worker work together and what GW's function is should be put front and center in a revamp of GW docs.

kotfic:
@jeffbaumes when decorating a function and turning it into a task object,  what you've really done is given the function two purposes in life. The first purpose is to execute the body of the function. The second purpose is to act as a factory for producing messages. When you use the traditional method,  you skip the factory entirely and just put a message on the queue using ```app.send_message(...)```.  This means task specific configuration that effects the message behavior between when you call a factory method ```.delay(...)``` and when you put the message on the queue is lost (for example rate limiting the message production).  

What's more,  configurations that are passed directly to ```app.send_message(...)``` must be defined separately on the producer side (Girder) and the consumer side (Girder Worker).  This means task configurations are being stored in two separate locations, if they every go out of sync we will start seeing undefined, difficult to debug, behavior. That isn't an insurmountable problem,  but *it is a problem that Celery has already solved*. 

Synchronization is critical in this context (and just generally when dealing with distributed systems like this).  There are potentially other methods that could be used to synchronize task and application configurations (e.g. REST endpoints), but what Celery has chosen is the *least common denominator* for syncing across producer and consumer processes: an importable package.

Importable packages tightly bind code and configuration,  which at first glance may seem bad. In fact this is critical because the configuration is fundamentally about how your code will be executed. Maintaining those in the same place reduces complexity and makes synchronization errors less likely. An importable package is also a pre-requisite for at least the consumer side of the equation (so the task can actually execute its function body). Importable packages prevent having to run a specific service to manage configuration,  or sync configuration files across systems,  they provide a versioned environment that ensures consistent interpretations of configuration values. Importable packages have strong tooling around their installation and deployment (pip) and once they are installed they have 100% up time. 

Ultimately, the user need that is precipitating all these changes is the desire to leverage Celery's [canvas](http://docs.celeryproject.org/en/latest/userguide/canvas.html) features. This allows chaining, chunking, grouping and map-reducing tasks in a nice, intuitive syntax. The issue with using canvas features is *Girder is no longer the only producer*.  When one task is chained onto another,  the first task is produced inside girder,  it is executed and its return value is passed on as input to the next task.  The next task is produced from inside that Girder Worker process. WIth canvas features,  Consumers are also Producers!  If the application and the task configurations are different between Girder and Girder Worker this could have disastrous consequences,  where-by  the order of a tasks execution in its workflow effects the result of the individual task.

kotfic:
sorry, realized i didn't answer your question about long term maintenance of the traditional method. The traditional method as it exists right now works by chance,  not by design. I don't see a problem with the traditional method continuing to exist as is, but i don't think its wise to put resources into supporting the traditional method if it fails to meet an individual's needs. Inevitably we'll end up solving problems to support the traditional method,  that are already solved in celery.  

In any case when you strip away the plugin related code, the REST endpoints and everything else,  the traditional method boils down to:

```python

app = celery.Celery('girder_worker', 
    backend='amqp://guest@localhost/'', 
    broker='amqp://guest@localhost/'')

app.send_task(
    'my_custom_package.tasks.my_complex_task', 
    job['args'], job['kwargs'], 
    queue=job.get('celeryQueue'), 
    headers={
             # custom configurations here
    })
```

If someone really needs to avoid importing either girder_worker or its tasks then they can still use this method (though with all the problems I've mentioned).

jeffbaumes:
Thanks @kotfic. Good to know the old and new methods can coexist, even if we envision transitioning away from the old approach.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+Currently the Worker plugin supports two ```otherFields```:
+
++ **celeryTaskName** should be the name of the task object. This is almost always the python importable path to the task as a string (unless you have specifically set the [name](http://docs.celeryproject.org/en/latest/userguide/tasks.html#names) when decorating the function). In our example this is the string "my_custom_package.tasks.fibonacci". If you do not specify a ```celeryTaskName``` in ```otherFields``` it will use "girder_worker.run" by default.
++ **celeryQueue** allows you to place the produces message on a different RabbitMQ queue than the default ```celery``` queue.
```
mgrauer: `place the produces message`?  ah, maybe `place the produced message` ?

mgrauer: why might you want to do this, place it on a different queue?  or will that be covered later?

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+For the final part of our example we use the jobModel singleton to save the job to the database and then schedule the job to execute.
```
mgrauer: This is a little hard to read as it jumps back to the "narrative" of the code block after delving into the details of the params.  Maybe better to remove this line and add it to the "narrative", then focus on the params to `createJob`.

```
In the above code we get the job model singleton which uses a factory method createJob(...) to create the job, then saves the job and schedules the job to execute.
```

Though I think all of that is pretty easily inferred from the code and comments and could be cut out.

You could start with something like

```
The params to the job model singleton `createJob` method are 
```

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+Currently the Worker plugin supports two ```otherFields```:
+
++ **celeryTaskName** should be the name of the task object. This is almost always the python importable path to the task as a string (unless you have specifically set the [name](http://docs.celeryproject.org/en/latest/userguide/tasks.html#names) when decorating the function). In our example this is the string "my_custom_package.tasks.fibonacci". If you do not specify a ```celeryTaskName``` in ```otherFields``` it will use "girder_worker.run" by default.
```
mgrauer: Should you give more background on `girder_worker.run` or its relationship to the traditional method?  You mention it here but it lacks context.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+### Notes about the Traditional method
+
+In general the traditional method is not the preferred method for producing tasks from Girder. As we will see, it takes several additional steps compared to the "Celery" method. More importantly, the traditional method can ***only*** use the default Celery [configurations](http://docs.celeryproject.org/en/latest/userguide/configuration.html) with the exception of the backend, and broker settings (for why, see "Issues with the Traditional Method" under the architectural notes section). 
+
+The one advantage of the traditional approach is that it does not require the package that defines your task objects to be installed in the Producer's environment (e.g. ```my_custom_package``` in our example does not have to be installed on the same server as Girder). This can be helpful if for some reason you do not have deploy time control over the Girder environment,  or if you have a particularly difficult set of dependencies that would need to be installed in both the Girder and Girder Worker Python environments (though,  See: "What if my task has complex install dependencies?" for suggestions on how to mitigate this issue with the Celery method).
```
mgrauer: This is kind of a big deal, in an imagined worker farm, where you could bring up new workers easily that have a new analysis, but to use them you'd need to bring down and back up all of your Girder instances.  Are there ways to make this lighter weight and not require all the dependencies in the Girder instance?

girder_worker/app.py
```diff
@@ -109,35 +126,54 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                **{'title': headers.pop('girder_job_title', Task._girder_job_title),
+                   'type': headers.pop('girder_job_type', Task._girder_job_type),
+                   'handler': headers.pop('girder_job_handler', Task._girder_job_handler),
+                   'public': headers.pop('girder_job_public', Task._girder_job_public),
                    'user': user,
                    'args': task_args,
                    'kwargs': task_kwargs,
                    'otherFields': dict(celeryTaskId=headers['id'],
-                                       **headers.get('girder_job_other_fields',
+                                       **headers.pop('girder_job_other_fields',
```
danlamanna: It's confusing that we initially make a distinction between options and headers (in the declaration), then we conflate the 2 into a single dictionary named headers, and pop all of the options out of headers to only have the real headers remain. I think we should just have 2 dictionaries.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pymongo==3.2.2
 pytz==2016.4
 requests[security]==2.10.0
 six==1.10.0
+girder-client==2.2.1
```
danlamanna: We also need to require `girder-client` in setup.py.

tests/integration/integration_test_endpoints/server/__init__.py
```diff
@@ -112,7 +143,7 @@ def test_celery_task_delay_fails(self, params):
     @describeRoute(
         Description('Test celery task apply_async'))
     def test_celery_task_apply_async(self, params):
-        result = fibonacci.apply_async((20,))
+        result = fibonacci.apply_async((20,), {})
```
danlamanna: Was this change left in by accident?

tests/integration/integration_test_endpoints/server/__init__.py
```diff
@@ -159,6 +190,30 @@ def test_celery_task_signature_apply_async_fails(self, params):
         result = signature.apply_async()
         return result.job
 
+    @access.token
+    @filtermodel(model='job', plugin='jobs')
+    @describeRoute(
+        Description('Test girder client is generated and can request scoped endpoints'))
+    def test_celery_girder_client_generation(self, params):
+
```
danlamanna: whitespace

tests/integration/integration_test_endpoints/server/__init__.py
```diff
@@ -159,6 +190,30 @@ def test_celery_task_signature_apply_async_fails(self, params):
...
+            'admin', girder_client_token=str(token['_id']))
+
+        return result.job
+
+    @access.token
+    @filtermodel(model='job', plugin='jobs')
+    @describeRoute(
+        Description('Test girder client with no token can\'t access protected resources'))
+    def test_celery_girder_client_bad_token_fails(self, params):
+
```
danlamanna: whitespace

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+@app.task
+def fibonacci(n):
+    if n == 1 or n == 2:
+        return 1
+    return fibonacci(n-1) + fibonacci(n-2)
+	
+```
+
+Each plugin must define a plugin class the inherits from ```girder_worker.GirderWorkerPluginABC```. GirderWorkerPluginABC's interface is simple. The class must define an \_\_init\_\_ function and a task\_imports function. \_\_init\_\_ takes the girder\_worker's celery application as its first argument. This allows the plugin to store a reference to the application, or change configurations of the application as necessary. The task\_imports function takes no arguments and must return a list of the package paths (e.g. importable strings) that contain the plugin’s tasks. As an example:
```
mgrauer: is `GirderWorkerPluginABC` a placeholder or a real name?  if a real name, it needs some explanation.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+In this example the producer side REST endpoint creates a token with the scope of the current user, then passes the REST argument ```file_id``` and the token to the task object's ```.delay(...)``` function.  On the consumer side,  the task has access to the GirderClient instance through ```self.girder_client```. To get access to ```self``` the task must be decorated with ```@app.task(bind=True)``` (See [bound tasks](http://docs.celeryproject.org/en/latest/userguide/tasks.html#bound-tasks) for more information). 
+
+# How do I write a custom Celery task for Girder Worker?
+
+*Note that this is taken almost directly from [Writing your own plugin](http://girder-worker.readthedocs.io/en/latest/developer-docs.html#writing-your-own-plugin)*
+
+Adding additional tasks to the girder\_worker infrastructure is easy and takes three steps. 
+
+1. Creating tasks
+2. Creating a plugin class
```
mgrauer: why does it have to be a plugin?  i'd want more background on why it is a plugin and a plugin system of what?

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+        # Update the celery application's configuration
+        # it is not necessary to change the application configuration
+        # this is simply included to illustrate that it is possible.
+        self.app.config.update({
+            'TASK_TIME_LIMIT': 300
+        })
+
+    def task_imports(self):
+        return ['gwexample.analyses.tasks']
```
mgrauer: will you actually show the gwexample.analyses.tasks implementation somewhere?  also, where does this string come from?

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+		
+```
+
+Finally, in order to make the plugin class discoverable, each plugin must define a custom entry point in its setup.py. For our example, this entry point looks like this:
```
mgrauer: discoverable to what, or by what?

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+      entry_points={
+          'girder_worker_plugins': [
+              'gwexample = gwexample:GWExamplePlugin',
+          ]
+      },
+      # ....
+      )
+```
+
+Python Entry Points are a way for python packages to advertise classes and objects to other installed packages. Entry points are defined in the following way:
```
mgrauer: might be better to move this up, in between

Finally, in order to make the plugin class discoverable, each plugin must define a custom entry point in its setup.py.

and 

For our example, this entry point looks like this:

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+```python
+entry_points={
+    'entry_point_group_id': [
+        'entry_point_name = importable.package.name:class_or_object',
+    ]
+}
+```
+
+The girder\_worker package introduces a new entry point group girder\_worker\_plugins. This is followed by a list of strings which are parsed by setuptools. The strings must be in the form name = module:plugin\_class Where name is an arbitrary string (by convention the name of the plugin), module is the importable path to the module containing the plugin class, and plugin_class is a class that inherits from GirderWorkerPluginABC.
```
mgrauer: what is an entry point group?

mgrauer: would be good to say when it is parsed by setuptools, i.e. `are parsed by setuptools at the time the package is pip installed`. or whatever is the right thing to say.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+One of the few disadvantages of the Celery method is requiring that the package that contains your task be installed on both the consumer (where the task is run) and the producer (where the task is enqueued). Normally this isn't a big problem, especially if your tasks are only using functions from Python's standard library. 
+
+In some cases however, tasks have complex dependencies that are difficult to install and it may be burdensome to install them in the producer (Girder) environment. Luckily this issue can be easily ameliorated by guarding imports in your task packages:
+
+```python
+from girder_worker.app import app
+
+try:
+	from complex_dependency import complex_function
```
mgrauer: indentation consistency

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+except ImportError:
+	pass
```
mgrauer: indentation consistency

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+    }
+	...
+)
+```
+
+This means the producer side code can be installed with pip as ```my_custom_package``` and the consumer side code (that includes the complex dependencies) can be installed as ```my_custom_package[consumer]```
+
+## Communicating with Girder from inside Tasks
+
+Tasks decorated with the Girder Worker application object have two main channels of interacting with the Girder instance that produced the task.
```
mgrauer: when you say `Girder instance`, you mean something like `set of Girder instances tied to the same database` or maybe `Girder host`.  Pedantically, the way you saying isn't correct in a multi-app server Girder environment.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
...tils.JobManager```. The job\_manager is available on the task object,  which means you must decorate the task with bind equal to true (e.g. ```@app.task(bind=true)```). This means the first argument to the task will be ```self``` and the job\_manger will be available as ```self.job_manager```.  For the most part,  task state transition is handled seamlessly through Girder Worker's use of [Celery Signals](http://docs.celeryproject.org/en/latest/userguide/signals.html). This means basic state transition (e.g.  running, success, failure, retry, etc) will be handled without any intervention within the task.  If you wish to use custom states,  update the Job progress, or to log specific information within your task,  you can do so using the this job\_manager. The job_manager is automatically generated regardless of whether you use the Traditional method or the Celery method of producing your task. The job\_manager is scoped so that it only has push access to the specific job it is assigned. If your task needs to leverage other components of the Girder REST API, including pushing data back to Girder or modifying Girder collections via the API,  then you should use the ```girder_client```
```
mgrauer: The job\_manager is available on the task object,  which means you must decorate the task with bind equal to true (e.g. ```@app.task(bind=true)```)

==>

The job\_manager is available on the task object,  which means you must decorate the task with bind equal to true in order to access the job\_manager (e.g. ```@app.task(bind=true)```)

mgrauer: For the most part,  task state transition is handled seamlessly through Girder Worker's use of [Celery Signals]

What does this mean?  that Celery will take care of changing the task state?  that Celery will take care of reflecting the task state back into the Job model?  both?

mgrauer: `basic state transition will be handled without any intervention within the task. ` . to me this implies all in Celery, is that what you mean or do you mean in the Job?

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+The second channel for interacting with Girder from within a task is through the ```girder_client``` attribute.  Like the ```job_manager``` it is available on the Celery task object (so you must set bind equal to true to access it). The girder\_client is an instance of ```girder_client.GirderClient```. It provides an authenticated session for making REST requests against the Girder API. The authentication is handled through a token provided to the task through the reserved keyword ```girder_client_token```. This will allow the girder\_client within the task to act with the same permissions granted to the token. By default the girder\_client\_token is None,  meaning the girder\_client instance within the task will have the permissions of an unauthenticated user.
```
mgrauer: `By default the girder\_client\_token is None` does this mean if you haven't set the token through the girder_client_token keyword arg?  If so, please clarify this.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+## Celery Hooks and the reserved keywords
+TODO
+
+## The jobInfoSpec and the task job\_manager
+TODO
+
+
+## Issues with the Traditional Method
+The fundamental issue with the Traditional Method is how does one synchronize Celery's configuration across producer (Girder) and consumer (Girder Worker) processes, keeping in mind that these processes may be taking place on two (or more) different physical machines. Traditionally Celery solves this problem through the Python import mechanism. The Celery application is configured as apart of the python package in which it is defined. By importing the application you get the same configuration on the producer as on the consumer. Assuming you are using the same version on both producer and consumer, then application is imported consumer-side when running ```girder-worker``` and is imported implicitly producer-side when importing a Celery Task object (e.g. ```from my_custom_package.tasks import fibonacci```).
```
mgrauer: s/apart/a part/

mgrauer: s/then application/then the application/

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
...nction it uses the task "name" (usually the string-ified importable path to the task). The advantage here is that the task does not need to be importable on the producer-side system, it doesn't even have to be installed at all. However,  because the task object itself is not producing the message all producer-side task level configuration is lost when using ```Celery.send_task(...)```. This might not be a deal-breaker except that the application object on the Girder producer-side is a *different application object* than on the Girder Worker consumer-side. The worker plugin does not import the ```girder_worker.app``` application object, instead it fabricates a new ```Celery(...)``` object before calling ```.send_task(...)```. While on the one hand this means the girder_worker package doesn't need to be installed in the girder environment it *also* means that the Celery application object on the producer-side is a differently configured object then on the consumer-side. Maybe that isn't a problem for you and your use-case (Celery has excellent defaults), but if you need more sophisticated configurations then currently the "Celery" method for remote task execution is your only option.
```
mgrauer: it uses the celery.send_task method which is set on the celery application object.  what does `set` mean here?

s/send_taks/send_task/

Maybe that isn't a problem ... can you give an example or two of why this would be a problem?

mgrauer:
I'm ok unblocking this PR on QA of the book.  I'm done my first pass now, and as long as the comments are responded to somewhere, I don't need them fixed here.  But I don't want all that QA work to have to be repeated.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+### Vocabulary
+
+Vocabulary can be an area of confusion when dealing with these systems. In an attempt to be clear we will draw a strong, jargon-like, distinction between the terms *Job* and *Task*.
+
+The term **Job** will always refer to the concept of a Girder [Job](http://girder.readthedocs.io/en/latest/plugins.html#jobs). Jobs are displayed on the Job's panel in Girder. They are represented in the database as [Job Models](https://github.com/girder/girder/blob/master/plugins/jobs/server/models/job.py). Girder Job's have various states and notification streams, and can be scheduled to execute on Girder Worker. 
+
+The term **Task** on the other hand will always refer to the Celery concept of a [Task](http://docs.celeryproject.org/en/latest/userguide/tasks.html). Tasks are typified by decorated python functions that can be scheduled to run on Girder Worker via the Celery [task calling api](http://docs.celeryproject.org/en/latest/userguide/calling.html). They are not persisted in any database but exist as python objects which are imported from a package.
```
kotfic: I'm adding some extra text here,  but just to clarify - I'm talking specifically about the Celery concept of a Task. This means a function that is decorated by ```@app.task```. This turns the function into a Python object (referred to as the 'task object'). That means the task doesn't exist until the python module it is apart of is executed (e.g. on import).

tests/integration/HACKING.md
kotfic: the only reason you would want to do this is because you are using an advanced feature of Celery/RabbitMQ.  Most people won't need this,  but i didn't want to leave it out entirely

tests/integration/HACKING.md
kotfic: In fact you would not need to bring down the girder instances,   only install the package in the girder instance's python environment.  For custom Celery tasks the Worker's will have to be restarted (to pick up the new tasks) - this should be primarily mitigated by the fact that many "user generated" tasks will be run with the generic 'docker run' task

girder_worker/app.py
kotfic: I'll admit that the code here is a little confusing,  but it is in service of a readable clean API for developers:

```python
fibonacci.delay(20,  girder_client_token="....")
```
vs.

```python
fibonacci.delay(20,  girder_headers={ "girder_client_token": "...." })
```

What's more this makes the distinction between kwargs that are used to control behavior in the ```before_task_publish``` signal vs options that are passed through to the message headers.  This means you would need a dict for ```girder_headers``` and a dict for ```girder_options```. This places unnecessary burden on the API user to know the difference between these things *or* means you've got to parse it out for them inside Task class. 

Each option that is possible has to be handled inside the Task code anyways to provide sensible defaults (in the case of kwargs being passed headers) or to provide functionality (in the case of kwargs effecting signal functionality).

requirements.txt
kotfic: this will get picked up in setup.py because of (regrettable) [magic](https://github.com/girder/girder_worker/blob/master/setup.py#L28-L71)

kotfic:
@danlamanna @mgrauer I am working on resolving these issues now. I'm hoping we can get this in,  then iteratively improve the documentation in PR's that don't block code improvements.

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+
+
+### Notes about the Celery Method
+
+While the Celery method is much more terse than the Traditional method,  the astute reader will recognize that it does not (as illustrated) provide the ability to modify the properties of the Girder job that is generated. In fact, using the fibonacci task introduced previously,  with the Celery method will produce a job in the Girder Job panel with the title "\<unnammed job\>". To remedy this situation a series of "special" or "reserved" Girder Worker specific keywords may be passed to any of the Celery Task API methods.  These keywords allow per-task invocation modification of the generated girder Job. They are:
+
++ **girder_job_title** This changes the job title, by default (e.g if girder\_job\_title is omitted) it is "\<unnamed job\>".
```
kotfic: https://github.com/girder/girder_worker/issues/168

tests/integration/HACKING.md
```diff
@@ -0,0 +1,468 @@
...
+#### Job Defaults
+
+The above keyword arguments can be used to set per-invocation values for the generated job model. While this is useful for specific invocations, generally Girder Job metadata will be similar across task invocations.  Default values for this metadata may be set by decorating the task definition with the ```girder_job``` decorator:
+
+```python
+from girder_worker.utils import girder_job
+from girder_worker.app import app
+
+
+@girder_job(title='Fibonacci Job',
```
kotfic: No unfortunately order is important here. I have added a line to the docs to make this clear.

tests/integration/HACKING.md
kotfic: ABC => Abstract Base Class - I have added a line that i hope makes this more clear.

tests/integration/HACKING.md
kotfic: @danlamanna  brings up a good point,  'Girder deployment'  is probably the right terminology

tests/integration/HACKING.md
mgrauer: I'm specifically thinking of the ElasticBeanstalk environment, where I'd expect in this case you'd need to redeploy Girder because of the additional package installs.

@danlamanna can correct me here.

kotfic: @mgrauer I talked a little with @danlamanna and this is true,  you would need to re-deploy girder.  But you would need to at least restart girder anyways to include whatever plugin or code actually imported the new task object and used it.  

I don't see this as being on the critical path,  because we will have to have a clean, frictionless workflow for deploying updates to the Girder Pool in ElasticBeanstalk environments regardless (e.g. for security updates,  critical bug fixes, etc).

PR #165, opened 2017-07-06
Add circleci support to run the integration tests

dorukozturk:
Runs the integration tests on CircleCi. We need to rebuild girder/girder:gw_integration_test and girder/girder_worker:gw_integration_test images and push them to Docker Hub for this to work.
@kotfic PTAL

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,38 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: python:2.7.13
+    working_directory: /root/girder_worker
+    steps:
+      - checkout
+      - run:
+          name: Install Docker client
+          command: |
+            set -x
+            VER="17.03.0-ce"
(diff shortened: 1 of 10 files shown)


kotfic:
With the move to [docker volume containers](https://docs.docker.com/engine/tutorials/dockervolumes/#creating-and-mounting-a-data-volume-container) I believe [this](https://github.com/dorukozturk/girder_worker/blob/circleci/tests/integration/common_tasks/common_tasks/__init__.py#L8-L22) is no longer true. Can you try removing it? if it breaks, you may need to add to the docker container volume Dockerfile scripts to change worker.dist.cfg to have a broker_url of 'amqp://guest:guest@rabbit/'

dorukozturk:
@kotfic PTAL. We are modifying worker.dist.cfg through the gw_entrypoint.sh [here](https://github.com/girder/girder_worker/pull/165/files#diff-04d4530f95b68218293dd64e8640897cR22).

kotfic:
This LGTM,  I am building/pushing the new images now

Issue #163, opened 2017-06-29
Clean up entrypoints/package config/Girder-Worker config


mgrauer:
This isn't related to the celery refactor necessarily, but would be good to be done after all of that churn settles down.

Additionally, making it so that pip installing girder-worker also sets the girder-worker config settings with the plugins.  This second step has bitten me and at least two others; it would be nice if the install process could be streamlined.

kotfic:
This isn't really related to the refactor,  I am going to remove it from the milestone

kotfic:
@mgrauer seems like there are a few different thing going on here?  Can we break this out into individual issues?

mgrauer:
Yes certainly. I can't recall what the issue title is about, perhaps I was transcribing it from someone who remembers the details better?

I am interested in having the girder worker config settings include the plugins when they are passed to pip, if that is possible, so that installation won't require this additional step. I.e., when I pip install GW with certain plugins, those plugins should be included in the GW config file.

kotfic:
Ah,  in that case I will mark this as Legacy,  and if you or anyone else remembers what the other components of the issues are they can raise new ones.

PR #166, opened 2017-07-11
Split entrypoint logic into a new module

jbeezley:
This is necessary for some of the item_tasks auto-discovery logic that I'm working on in a separate branch.  In particular, it provides a new function that will import all task modules from plugins which will serve to register all decorated tasks as a side effect.


mgrauer:
Did this PR help move towards or provide a basis for #163 ?

jbeezley:
It's partially related, but there is more work to be done especially in regards to the config file.

PR #169, opened 2017-08-03
Use pytest hooks to magically output response content

cjh1:


cjh1:
I  am aware this may be a little too magical!

danlamanna:
I think this might be too confusing to add to GW. Even though it's more
work/boilerplate, I think it might be better to just add
r.response.content as our assert message each time.

cjh1:
Not sure its any less confusing than what pytest already does with asserts :smiley: My only concern with the manual approach is ensuring that developers do it, but we can add that as a best practice or something.

danlamanna:
I agree, but the main difference I see is that they write, test, and
maintain that code and we only ever have to deal with the public facing
interface.

Does anyone else want to weigh in, @kotfic?

zachmullen:
I actually think this is nice, but my opinion here doesn't matter since I have not worked with this part of the code.

cjh1:
Apparently no one believes in magic!

PR #175, opened 2017-08-23
Various ansible improvements

danlamanna:
- Stop using lookup pipe for finding executables  
- Stop crashing when girder_worker_pypi_version isn't passed  
- Call daemon-reload after changing service file ansible role  
- Make configure.yml run as root when no virtualenv is defined


devops/ansible/roles/girder-worker/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 girder_worker_install_source: "pypi" # must be one of pypi, git
 
 # versions to install, pypi will default to latest
-girder_worker_pypi_version: null
+# girder_worker_pypi_version: null
```
opadron: Should remove this commented line.

PR #176, opened 2017-08-23
CentOS support for Ansible role

danlamanna:
~~Depends on #175, actual diff is https://github.com/girder/girder_worker/compare/various-ansible-improvements...centos-support.~~


devops/ansible/roles/girder-worker/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 girder_worker_install_source: "pypi" # must be one of pypi, git
 
 # versions to install, pypi will default to latest
-girder_worker_pypi_version: null
+# girder_worker_pypi_version: null
```
mgrauer: cleanup?

PR #179, opened 2017-08-25
Stop trying to fetch unshallow objects for galaxy deploy

danlamanna:
It appears CircleCI 2 must do a complete clone instead of a shallow
one like CircleCI 1.


dorukozturk:
LGTM

PR #183, opened 2017-09-01
Wait for job to be in running state before revoking

cjh1:
Its possible for the job to be revoked before its moves into the running state depending on how quickly things are running. Add a wait for the running state to remove this timing window.


cjh1:
@danlamanna This should resolve the intermittent failure you were running into.

PR #188, opened 2017-09-07
Ensure _revoked_tasks always returns a list (even if empty)

kotfic:
Celery's invoke.revoked(...) function may return None instead of a
list. Calling .get() on _worker_inspector may raise an AttributeError
because of this.  This ensures if None is returned _revoked_tasks
still returns an empty list.

@cjh1 PTAL
@brianhelba I believe this should resolve your issue

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -349,7 +349,12 @@ def _worker_inspector(task):
 
 # Get this list of currently revoked tasks for this worker
 def _revoked_tasks(task):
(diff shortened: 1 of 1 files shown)


brianhelba:
This resolves my issue!

Thanks so much for the quick fix, @kotfic and @cjh1 .

cjh1:
Looks like the reason this was not caught by our integration test is @brianhelba is using the MongoDB backends, so its probably a different code path.

PR #190, opened 2017-09-09
Celery chaining

### LINKED ISSUE #140: Support chaining celery tasks
### Deliverables:
+ Demonstrate syntax like the following from girder REST endpoint:
```python
	(task1.s () | task2.s()).delay()
```
+ Demonstrate asynchronously calling a girder worker task from another girder worker task (e.g. like taskflow - allows for conditionals etc). 

### Technical Notes:
+ Note on dependencies: 
   + We need a REST endpoint for creating child jobs because in the case of a chain,  the second task is actually “produced” from a girder_worker instance. 
    + Parent child relationships need to exist inside girder because otherwise there is no way to visually relate chains of tasks
    + Celery 3 has no clear way to relate one task to another.  To allow chaining we need Celery 4 to be the default

### Depends on:
+ Creation of girder job models via REST endpoints in worker plugin (https://github.com/girder/girder/issues/2018)
+ Allow parent-child relationships between girder job models (https://github.com/girder/girder/issues/2017)
+ Clean up Celery 3 & 4 support in app.py (https://github.com/girder/girder_worker/issues/139)

dorukozturk:
Fixes #140 
Depends on girder/girder/pull/2335

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -7,6 +7,7 @@ import traceback as tb
 from distutils.version import LooseVersion
 import celery
 from celery import Celery, __version__
+import requests
 
 from celery.result import AsyncResult
 from celery.signals import (
@@ -25,6 +26,7 @@ import jsonpickle
 from kombu.serialization import register
 import six
 from .utils import JobStatus, StateTransitionException
+import json
 
 
 class GirderAsyncResult(AsyncResult):
@@ -75,8 +77,7 @@ class Task(celery.Task):
     reserved_headers = [
         'girder_client_token',
         'girder_api_url',
-        'girder_result_hooks',
-        'girder_parent_id']
+        'girder_result_hooks']
 
     # These keys will be available in the 'properties' dictionary inside
     # girder_before_task_publish() but will not be passed along in the message
@@ -188,14 +189,30 @@ def _maybe_model_repr(obj):
     return obj
 
 
-def _set_arg_to_chained_tasks(body, arg, val):
-    # Helper function which sets a value to chained child tasks
-    chained_tasks = body[2].get('chain')
-    if chained_tasks:
-        for i in chained_tasks:
-            if arg not in i.kwargs.keys():
-                    i.kwargs[arg] = val
-@before_task_publish.connect
+def _get_token(user=None):
+    """Helper function to get the token"""
+    from girder.utility.model_importer import ModelImporter
+    from girder.api.rest import getCurrentUser
+    token_model = ModelImporter.model('token')
+    scope = 'jobs.rest.create_job'
+    if user:
+        token = token_model.createToken(scope=scope, user=user)
+    else:
+        token = token_model.createToken(scope=scope, user=getCurrentUser())
+
+    return token
+
+
+def _get_attr_from_task(task, attr):
+    """Tries to get a given attribute from a task"""
+    request = task.request
+    if hasattr(request, attr):
+        return getattr(request, attr)
+    else:
+        logger.warn('Failed to get {} from parent task in the worker context'.format(attr))
+
+
+@before_task_publish.connect  # noqa: C901
 def girder_before_task_publish(sender=None, body=None, exchange=None,
                                routing_key=None, headers=None, properties=None,
                                declare=None, retry_policy=None, **kwargs):
@@ -240,37 +257,54 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
                 headers['jobInfoSpec'] = utils.jobInfoSpec(job)
 
             except ImportError:
-                # TODO: Check for self.job_manager to see if we have
-                #       tokens etc to contact girder and create a job model
-                #       we may be in a chain or a chord or some-such
-                pass
+                # Current task will be the parent task in chain case
+                parent_task = app.current_task
+                gc = GirderClient(apiUrl=_get_attr_from_task(parent_task, 'girder_api_url'))
+                gc.token = _get_attr_from_task(parent_task, 'girder_client_token')
+
+                task_args = tuple(_walk_obj(body[0], _maybe_model_repr))
+                task_kwargs = _walk_obj(body[1], _maybe_model_repr)
+                parameters = {
+                    'title': headers.pop('girder_job_title', Task._girder_job_title),
+                    'type': headers.pop('girder_job_type', Task._girder_job_type),
+                    'handler': headers.pop('girder_job_handler', Task._girder_job_handler),
+                    'public': headers.pop('girder_job_public', Task._girder_job_public),
+                    'args': json.dumps(task_args),
+                    'kwargs': task_kwargs,
+                    'otherFields': json.dumps(
+                        dict(celeryTaskId=headers['id'],
+                             celeryParentTaskId=_get_attr_from_task(parent_task, 'id'),
+                             **headers.pop('girder_job_other_fields',
+                                           Task._girder_job_other_fields)))
+                }
+
+                try:
+                    response = gc.post('job', parameters=parameters, jsonResp=False)
+                except requests.exceptions.RequestException as e:
+                    logger.warn('Failed to post job: {}'.format(e))
+
+                if response.ok:
+                    headers['jobInfoSpec'] = response.json().get('jobInfoSpec')
 
         if 'girder_api_url' not in headers:
             try:
                 from girder.plugins.worker import utils
                 headers['girder_api_url'] = utils.getWorkerApiUrl()
-                _set_arg_to_chained_tasks(body, 'girder_api_url', headers['girder_api_url'])
             except ImportError:
-                # TODO: handle situation where girder_worker is producing
-                #       the message Note - this may not come up at all
-                #       depending on how we pass girder_api_url through to
-                #       the next task (e.g. in the context of chaining
-                #       events)
-                pass
+                parent_task = app.current_task
+                headers['girder_api_url'] = _get_attr_from_task(parent_task, 'girder_api_url')
 
         if 'girder_client_token' not in headers:
             try:
-                from girder.utility.model_importer import ModelImporter
-                headers['girder_client_token'] = \
-                    ModelImporter.model('token').createToken()
-                _set_arg_to_chained_tasks(body, 'girder_client_token', headers['girder_client_token'])
+                try:
+                    token = _get_token(user=user)
+                except NameError:
+                    token = _get_token()
+                headers['girder_client_token'] = token['_id']
             except ImportError:
-                # TODO: handle situation where girder_worker is producing
-                #       the message Note - this may not come up at all
-                #       depending on how we pass girder_token through to
-                #       the next task (e.g. in the context of chaining
-                #       events)
-                pass
+                parent_task = app.current_task
+                headers['girder_client_token'] = _get_attr_from_task(parent_task,
+                                                                     'girder_client_token')
 
         if 'girder_result_hooks' in headers:
             # Celery task headers are not automatically serialized by celery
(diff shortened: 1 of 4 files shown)


dorukozturk:
@kotfic This is ready for a review. PTAL.

girder_worker/app.py
```diff
@@ -130,10 +132,30 @@ def call_item_task(self, inputs, outputs={}):
         return self.run.call_item_task(inputs, outputs)
 
 
-@before_task_publish.connect
+@before_task_publish.connect # noqa
```
zachmullen: Why `# noqa`? Too complex?

dorukozturk: Unfortunately yes. I am planning to reduce the complexity of that function in another PR.

girder_worker/app.py
```diff
@@ -130,10 +132,30 @@ def call_item_task(self, inputs, outputs={}):
...
 def girder_before_task_publish(sender=None, body=None, exchange=None,
                                routing_key=None, headers=None, properties=None,
                                declare=None, retry_policy=None, **kwargs):
+
+    def _setArgToChainedTasks(arg, val):
+        # Helper function which sets a value to chained child tasks
+        if body[2].get('chain'):
+            for i in body[2]['chain']:
+                i.kwargs[arg] = val
```
kotfic: This should not overwrite kwargs[arg]  if arg in kwargs

girder_worker/app.py
```diff
@@ -161,19 +182,33 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
             headers['jobInfoSpec'] = utils.jobInfoSpec(job)
 
         except ImportError:
-            # TODO: Check for self.job_manager to see if we have
-            #       tokens etc to contact girder and create a job model
-            #       we may be in a chain or a chord or some-such
-            pass
+            gc = GirderClient(apiUrl=headers['girder_api_url'])
+            token = headers['girder_client_token']
+            gc.token = token
```
kotfic: just a style nitpick,  can you combined line 189 and 190

girder_worker/app.py
```diff
@@ -130,10 +132,30 @@ def call_item_task(self, inputs, outputs={}):
         return self.run.call_item_task(inputs, outputs)
 
 
-@before_task_publish.connect
+@before_task_publish.connect # noqa
 def girder_before_task_publish(sender=None, body=None, exchange=None,
                                routing_key=None, headers=None, properties=None,
                                declare=None, retry_policy=None, **kwargs):
+
+    def _setArgToChainedTasks(arg, val):
```
kotfic: eventually we will want this to be much more sophisticated (e.g. with chords and other chaining)  but this is probably fine for this PR

tests/task_signal_test.py
```diff
@@ -52,18 +52,15 @@ def setUp(self):
...
                       body=[(), {}, {}],
                       headers=self.headers)
 
         create_job = mi.model.return_value.createJob
-
         with mock_worker_plugin_utils():
             girder_before_task_publish(**inputs)
-
-        self.assertTrue(not gcu.called)
+        gcu.assert_called_once()
```
kotfic: Sorry,  can you explain why this needed to change?  Thanks!

kotfic:
Can we also test passing seperate reserved keywords etc to the second task in the chain (including different girder token etc)

tests/task_signal_test.py
dorukozturk: @kotfic Before the token we use was orphan, now we specify a user to the token. Thus getCurrentUser gets called so that we can attach it to the token.

kotfic:
@dorukozturk sorry to have let this slide,  can you rebase and fix conflicts? thanks!

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
                 headers['jobInfoSpec'] = utils.jobInfoSpec(job)
 
             except ImportError:
-                # TODO: Check for self.job_manager to see if we have
-                #       tokens etc to contact girder and create a job model
-                #       we may be in a chain or a chord or some-such
-                pass
+                gc = GirderClient(apiUrl=headers['girder_api_url'])
```
kotfic: Error if girder_api_url is not in headers.  shouldn't this be coming from task.request anyways?

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                gc.token = headers['girder_client_token']
```
kotfic: error if girder_client_token is not in headers. shouldn't this be coming from task.request anyways?

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                task_args, task_kwargs = body[0], body[1]
+                parameters = {
+                    'title': headers.pop('girder_job_title', Task._girder_job_title),
+                    'type': headers.pop('girder_job_type', Task._girder_job_type),
+                    'handler': headers.pop('girder_job_handler', Task._girder_job_handler),
+                    'public': headers.pop('girder_job_public', Task._girder_job_public),
+                    'args': json.dumps(task_args),
```
kotfic: shouldn't this call _mayber_repr_?

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                    'kwargs': task_kwargs,
```
kotfic: what if task_kwargs aren't serializable?

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                    'otherFields': json.dumps(dict(celeryTaskId=headers['id'],
+                                                   celeryParentTaskId=headers['parent_id'],
```
kotfic: what if parent_id is None?  shouldn't this be coming from task.request anyways?

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                                                   **headers.pop('girder_job_other_fields',
+                                                                 Task._girder_job_other_fields)))
+                    }
+                job = gc.post('job', parameters=parameters)
```
kotfic: What if this post fails?

girder_worker/app.py
```diff
@@ -231,15 +256,29 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                headers['jobInfoSpec'] = job['jobInfoSpec']
```
kotfic: What if jobInfoSpec is not in the returned Job?

tests/integration/integration_test_endpoints/server/raw.py
```diff
@@ -191,3 +194,34 @@ def test_celery_task_revoke_in_queue(self, params):
...
+        Description('Test chained celery tasks')
+    )
+    def test_celery_task_chained(self, params):
+        jobModel = ModelImporter.model('job', 'jobs')
+        user = self.getCurrentUser()
+        # F(F(F(6))) --> F(F(8)) --> F(21) --> 10946
+        result = (fibonacci.s(6) | fibonacci.s() | fibonacci.s()).delay()
+        result.wait(timeout=10)
+        job_1 = result.job
+        job_2 = jobModel.load(job_1['parentId'], user=user)
```
kotfic: Can we wrap this up inside a 'children()' method on the jobModel?

tests/integration/integration_test_endpoints/server/raw.py
```diff
@@ -191,3 +194,34 @@ def test_celery_task_revoke_in_queue(self, params):
...
+    @describeRoute(
+        Description('Test chained celery tasks with a bad token')
+    )
+    def test_celery_task_chained_bad_token_fails(self, params):
+        # By default the first task (fib(5)) should try to set it's token
+        # to the second task. If user passes a token, it should not overwrite
+        # the given token though.
+
+        result = (fibonacci.s(5) | fibonacci.si(5, girder_client_token='')).delay()
+        result.wait(timeout=10)
```
kotfic: This is going to slow down the whole test suite.  Is there a way to poll for this? Or just call result.get()?

tests/task_signal_test.py
```diff
@@ -63,7 +63,7 @@ def test_girder_before_task_publish_with_jobinfospec_no_job_created(self, gcu, m
         with mock_worker_plugin_utils():
             girder_before_task_publish(**inputs)
 
-        self.assertTrue(not gcu.called)
+        gcu.assert_called_once()
```
kotfic: Can you help me understand why this is being changed?

tests/integration/integration_test_endpoints/server/raw.py
```diff
@@ -191,3 +194,34 @@ def test_celery_task_revoke_in_queue(self, params):
...
+    @filtermodel(model='job', plugin='jobs')
+    @describeRoute(
+        Description('Test chained celery tasks with a bad token')
+    )
+    def test_celery_task_chained_bad_token_fails(self, params):
+        # By default the first task (fib(5)) should try to set it's token
+        # to the second task. If user passes a token, it should not overwrite
+        # the given token though.
+
+        result = (fibonacci.s(5) | fibonacci.si(5, girder_client_token='')).delay()
```
kotfic: Can we test this against a task that hits an endpoint that actually requires the authorization of a specific girder_client_token?

girder_worker/app.py
```diff
@@ -186,7 +188,30 @@ def _maybe_model_repr(obj):
     return obj
 
 
-@before_task_publish.connect
+def _set_arg_to_chained_tasks(body, arg, val):
```
kotfic: This is going to eventually need to be much more sophisticated,  we're going to need a function that actually walks the chain/group attributes and sets values.  That can be addressed in another PR, but maybe we should change the name here?  _set_arg_on_child_tasks?

dorukozturk:
@kotfic One more time this is ready for review.

girder_worker/app.py
```diff
@@ -232,35 +248,86 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
...
+                    assert 'girder_api_url' in parent_task.request, \
+                        "Parent task's request does not contain girder_api_url"
+
+                    assert 'girder_client_token' in parent_task.request, \
+                        "Parent task's request does not contain girder_client_token"
+
+                    assert 'id' in parent_task.request, \
+                        "Parent task's request does not contain id"
+
+                    assert 'id' in headers, 'id is not in headers'
```
zachmullen: These feel like misuse of `assert`. Can we change these to `if` blocks that raise exceptions?

danlamanna: +1

danlamanna: Since there was an in person debate about this, aside from being semantically incorrect - Python actually removes assertion statements when optimized, so `assert` definitely shouldn't be used to determine control flow.

dorukozturk: I changed it to use if statements and it raises a custom exception now. Do you guys think ParentAttributeException is a good name for the exception?

kotfic:
This block of exceptions are intended to test whether the necessary attributes for job creation are present. Maybe throwing a built in RuntimeWarning or a custom exception like “MissingJobArguments” that derives from RuntimeWarning?

cjh1:
:1st_place_medal:

mgrauer:
🥕 ⛓ 😻 

carrot was the closest I could find

dorukozturk:
End of an era ... :)

Issue #191, opened 2017-09-11
test_celery_task_revoke_in_queue still causing problems


kotfic:
When running all tests in parallel I am still seeing errors with ```test_celery_task_revoke_in_queue```.  

Specifically I am running ```pytest -n 20 -vlx --tb=long``` and intermittently seeing the following error:
 
```
[gw12] FAILED test_celery.py::test_celery_task_revoke_in_queue[admin] 

=================================================== FAILURES ====================================================
____________________________________ test_celery_task_revoke_in_queue[admin] ____________________________________
[gw12] linux2 -- Python 2.7.13 /home/kotfic/.venvs/gw2/bin/python2
session = <tests.integration.utilities.GirderSession object at 0x7fc0389485d0>

    def test_celery_task_revoke_in_queue(session):
        url = 'integration_tests/celery/test_task_revoke_in_queue'
        r = session.post(url)
        assert r.status_code == 200, r.content
    
        with session.wait_for_canceled(r.json()['_id']) as job:
>           assert [ts['status'] for ts in job['timestamps']] == \
                [JobStatus.CANCELED]
E           assert [2, 5] == [5]
E             At index 0 diff: 2 != 5
E             Left contains more items, first extra item: 5
E             Full diff:
E             - [2, 5]
E             + [5]

job        = {'_accessLevel': 2, '_id': '59b6bfa8236e0901b0c7a6aa', '_modelType': 'job', 'args': [], ...}
r          = <Response [200]>
session    = <tests.integration.utilities.GirderSession object at 0x7fc0389485d0>
ts         = {'status': 5, 'time': '2017-09-11T16:54:00.816000+00:00'}
url        = 'integration_tests/celery/test_task_revoke_in_queue'

test_celery.py:105: AssertionError
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================== 1 failed, 20 passed in 6.25 seconds ======================================

```

See: #183

PR #192, opened 2017-09-11
Block queue using cancelleable tasks

cjh1:
Rather than using a task that sleep use a cancellable task that can be cleaned up after the test case is run. This removes a small timing window that can cause tests to fail when run in parallel.


cjh1:
Fixes #191 ?

cjh1:
If we want preserve being able to run in parallel should we enable it on CI?

cjh1:
@kotfic Can you confirm this still fixes your issue?

PR #195, opened 2017-09-15
Centos multi vm

dorukozturk:
This PR is for adding CentOS as an option for the multi-vm example.


dorukozturk:
@danlamanna PTAL

devops/ansible/roles/redhat-rabbitmq/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+- name: Install erlang
+  yum:
+    name: erlang
+    state: present
+  become: yes
```
danlamanna: This line can be removed.

PR #199, opened 2017-09-22
Use direct paths when possible.

manthey:
If a direct_path parameter is given as part of a input specification, and that path is reachable, use it instead of downloading the file.  This only works for files where the parent is not being fetched.

This changes the girder_io plugin to use the direct path instead of download the file to the temp directory.

This changes the docker plugin to mount direct path files as their own read-only volumes.  This maintains the same level of isolation as before, except that the path name on the girder system is revealed to the worker.

See also https://github.com/girder/girder/pull/2356.

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -1,7 +1,7 @@
 .. _worker-plugins:
 
 Application Plugins
-=======
+===================
 
 The Girder Worker application plugin system is used to extend the core functionality of
 Girder Worker in a number of ways. Application plugins can execute any Python code when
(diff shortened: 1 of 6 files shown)


kotfic:
I am a big fan of this optimization,  but I am wondering if it should be directly encoded into the spec like this? 

The optimization really only makes sense in certain deployment contexts,  and so I would like to propose a different approach. Rather than https://github.com/girder/girder/pull/2356, create a girder item task setting like "Local Assetstore Optimizations."  When true direct_path is included in all input bindings of resource type 'file.'  On the girder worker side add a command line option (or configuration) to the effect of '--girder=local'   If true it will use the direct_path rather than downloading (after appropriate existence checks etc).

This moves the optimization out of the specification and makes it a deploy time concern.

manthey:
What do you mean by a "girder item task setting"?  It sounds like an item_tasks property, but my current use case doesn't (yet) use item_tasks.  Are we expecting to only support item_tasks in the future (i.e., the remote_worker plugin will depend on the item_tasks plugin)?

Although running girder_worker on the same filesystem as girder may not be the only use case, it is common enough that I would suggest that the default configuration should be to try to use the local path if provided, and the new option would be to never try (--girder=download).

zachmullen:
Maybe a worker plugin system setting, then?

PR #204, opened 2017-10-25
Refactor endpoints into separate resources

cjh1:
Group related endpoints in a separate resource and mount then under a common prefix.


requirements-dev.txt
```diff
@@ -3,6 +3,7 @@ coveralls==1.1
 flake8==3.3.0
 flake8-blind-except==0.1.1
 flake8-docstrings==1.1.0
+pydocstyle==2.0.0
```
kotfic: Can you file an issue related to the pydocstyle issue? Otherwise this looks good to me

PR #206, opened 2017-10-31
Fallback to a regular container if nvidia is unavailable.

manthey:
When running an nvidia container, if we can't connect to nvidia, run as a regular container.

Ideally, it would be good if we knew if nvidia was *required* or only *preferred*.  When it is just preferred, it is better to run without nvidia support then to not run at all.


manthey:
@zachmullen If there is a better way to handle this, I'd be happy to change it.  It works in my test of a docker job on a system without an nvidia GPU (running a task that doesn't actually benefit from the GPU).

PR #209, opened 2017-11-06
Fix Dockerfile so it works with pip-tools introduced in ec85660a

kotfic:
There are a few other small issues that had to be resolved,  including adding r-base to the required packages.


kotfic:
@danlamanna Could you review this and also set up an automated build on dockerhub.io?

kotfic:
@cjh1 for some of the tests,  that will be going away soon hopefully

PR #210, opened 2017-11-06
Task __call__ function with transform

kotfic:
This PR adds a __call__ function to the girder_worker.app.Task object which handles objects that have transform() functions defined.   These objects are created through the serialization/deserialization mechanism's made available in girder/girder_worker_utils

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -182,10 +182,9 @@ class Task(celery.Task):
 
 
 def _maybe_model_repr(obj):
-    try:
+    if hasattr(obj, '__repr_model_') and hasattr(obj._repr_model_, '__call__'):
         return obj._repr_model_()
-    except AttributeError:
-        return obj
+    return obj
 
 
 # TODO: any exceptions raised in this function are not raised
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -183,7 +183,7 @@ class Task(celery.Task):
 
 def _maybe_model_repr(obj):
     try:
-        return obj.model_repr()
+        return obj._repr_model_()
     except AttributeError:
         return obj
 
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -17,11 +17,11 @@ from .utils import JobStatus, StateTransitionException
 from girder_client import GirderClient
 
 from kombu.serialization import register
-from girder_worker_utils import json, _walk_obj
+from girder_worker_utils import _walk_obj
 
-
+import jsonpickle
 import functools
-from girder_worker_utils.json import object_hook
+# from girder_worker_utils.json import object_hook
 
 class GirderAsyncResult(AsyncResult):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


girder_worker/app.py
```diff
@@ -129,11 +136,57 @@ def describe(self):
...
+        try:
+            grh = self.request.girder_result_hooks[idx]
+            if hasattr(grh, 'transform') and hasattr(grh.transform, '__call__'):
+                return grh.transform(result)
+        except IndexError:
+            return result
+
+    def _maybe_transform_argument(self, arg):
+        if hasattr(arg, 'transform') and hasattr(arg.transform, '__call__'):
+            return arg.transform()
```
cjh1: How does girder_client aka 'gc' [reference here]( https://github.com/girder/girder_worker_utils/blob/master/girder_worker_utils/transforms/girder_io.py#L12) get passed in, is that still todo?

kotfic: On the one hand you can use the transform like this:

```python
mytask.delay(GirderFileId('58dd37438d777f0aef5d8c75'))
```
This will create a default girder client with your permissions. Alternatively you
you can instantiate your own girder client with different permissions and pass it in using the ```gc``` keyword:

```python
token = Token().createToken(scope='some.custom.scope')
girder_client = GirderClient(apiUrl=some_api_url)
girder_client.token = token

mytask.delay(GirderFileId('58dd37438d777f0aef5d8c75', gc=girder_client))
```

This isn't currently documented anywhere,  but is the intended functionality

cjh1: Thanks for explanation.

girder_worker/app.py
```diff
@@ -273,6 +355,12 @@ def gw_task_prerun(task=None, sender=None, task_id=None,
     except AttributeError:
         task.girder_client = None
 
+    # Deserialize girder_client_tokens if they exist
```
cjh1: This comment is a little misleading?

kotfic: see d1cee54

girder_worker/app.py
```diff
@@ -129,11 +143,58 @@ def describe(self):
...
+        finally:
+            _walk_obj(args, self._maybe_cleanup)
+            _walk_obj(kwargs, self._maybe_cleanup)
+
+
+def _maybe_model_repr(obj):
+    try:
+        return obj._repr_model_()
+    except AttributeError:
+        return obj
```
cjh1: We should probably explicitly checked for the existence of ```_repr_model```. Otherwise, an ```AttributeError``` raised in the users ```_repr_model``` implementation it might be hard to track down. This has bitten me before.

jbeezley:
Is there a plan for integration style testing for some of this transform behavior?

jbeezley:
I guess @cjh1 mentioned it.  I guess I'm the :snail: now.

cjh1:
My ```docker_run``` tests will exercise some of the transform code, but yes, we should have more comprehensive coverage in the future.

cjh1:
:sparkles:

PR #215, opened 2017-11-29
Fixes typo with _maybe_model_repr utility function

kotfic:


girder_worker/app.py
```diff
@@ -182,7 +182,7 @@ def __call__(self, *args, **kwargs):
 
 
 def _maybe_model_repr(obj):
-    if hasattr(obj, '__repr_model_') and hasattr(obj._repr_model_, '__call__'):
+    if hasattr(obj, '_repr_model_') and hasattr(obj._repr_model_, '__call__'):
```
zachmullen: `and callable(obj._repr_model_)`

kotfic: I wanted to avoid callable because it is deprecated (i.e. not available) in python 3.0 and 3.1. hasattr is technically [wrong](https://stackoverflow.com/a/18704793),  but it is consistently wrong across python versions

kotfic: will use https://pythonhosted.org/six/#six.callable

PR #216, opened 2017-12-01
Remove the entrypoint code's dependency on celery

kotfic:
This is in preparation for moving entrypoint and describe
functionality into girder/girder_worker_utils

### DIFF
--- a/girder_worker/entrypoint.py
+++ b/girder_worker/entrypoint.py
@@ -2,7 +2,11 @@ from importlib import import_module
 
 import six
 from stevedore import extension
-import celery
+
+try:
+    from celery import current_app
+except ImportError:
+    current_app = None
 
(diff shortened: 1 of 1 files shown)


cjh1:
Whats the motivation for moving the entrypoint into girder_work_utils?

kotfic:
@cjh1 i guess the idea is that discovery of installed girder_worker task plugins isn't exclusively a girder_worker run-time concern. i'd like to be able to write a command line runner that relies on click for instance that discovers installed tasks and provides a way to run them from the CLI.  In that mode of operation there wouldn't be a running instance of girder (or potentially even a running broker).  It's also possible these functions wouldn't be decorated as girder_worker tasks yet (in which case they would only be run as command line tools).

kotfic:
We can come back to this,  for now i'm going to close this

PR #218, opened 2017-12-01
Refactor integration tests

kotfic:
Refactor integration tests by using a Makefile to coordinate various docker/docker-compose commands.

PR #221, opened 2017-12-07
Refactor Celery config into celeryconfig.py

kotfic:
This also adds support for GIRDER_WORKER_BROKER and
GIRDER_WORKER_BACKEND environment variables which take precedence over
variables defined in worker.local.cfg and worker.dist.cfg


kotfic:
FWIW, this is blocking https://github.com/girder/girder_worker/pull/218

kotfic:
@cjh1 just wanted to ping you in case you had any input

PR #223, opened 2017-12-15
Fix running docker integration tests locally

kotfic:
I am still seeing failures locally that need to be resolved


cjh1:
@kotfic This should be good to go, if my changes look good.

girder_worker/docker/transforms/__init__.py
```diff
@@ -135,21 +136,22 @@ class TemporaryVolume(_TemporaryVolumeBase):
     A temporary volume can also be create in a particular host directory by providing the
     `host_dir` param.
     """
-    def __init__(self, host_dir=None):
+    def __init__(self, host_dir=None, mode=0775):
```
zachmullen: Should probably just use `0777` since the umask bits will be omitted automatically.

zachmullen: Looks like you beat me to it.

kotfic: The goal here is to make the directory read/writable by the user and the group.  I'm not quite sure i understand what you mean by the umask bits will be omitted automatically?

kotfic:
@cjh1 Those changes look good to me

girder_worker/docker/transforms/__init__.py
kotfic: This ```mode``` is being passed to a call to ```os.chmod``` specifically because sometimes we need to set the mode to 0777.  It isn't being passed to ```os.mkdir```

zachmullen: Sorry I thought this was `0755` not `0775`. Why is group write necessary? Might be a good place for a comment in the code.

kotfic:
@zachmullen PTAL

PR #225, opened 2017-12-18
Pull celery based tests into pytest, move legacy tests into girder_worker/core/

kotfic:
This PR pulls tests related to modern celery-style tasks out into tox and pytest.  It moves the traditional-style task tests (run with ctest) into girder_worker/core/tests.  Linting of legacy code is handled by ctest,  linting of modern code is handled by tox.  Coverage is produced by both test harnesses and tox generates a combine coverage report.

This PR prepares the code base to transition the celery-style code to be python 2/3 compatible while leaving the legacy code as python 2 only.


kotfic:
@danlamanna PTAL

@brianhelba Some of the choices wrt.  setup.cfg/tox.ini/coverage etc are arbitrary,  if you have some opinions/guidance that would be great.

tox.ini
```diff
@@ -0,0 +1,88 @@
...
+skip_install = true
+basepython = python2.7
+deps = coverage
+commands =
+  coverage combine
+  coverage report
+
+[testenv:lint]
+skip_install = true
+basepython = python2.7
```
danlamanna: In the spirit of making GW Py3 compliant, should this be left out? I noticed running against Python3 produced 5 linting errors that could be resolved.

.coveragerc.pytest
```diff
@@ -0,0 +1,8 @@
+[run]
```
danlamanna: Is there any reason this file can't be named `.coveragerc`?

tox.ini
```diff
@@ -0,0 +1,88 @@
...
+  --cov-report= \
+  {posargs}
+
+# This report is used to merge coverage reports from testenv and from
+# the legacy tests run by ctest. When just running 'tox' it will print
+# the pytest coverage report. When running ctest && tox it will print
+# the combine coverage.
+[testenv:coverage-report]
+skip_install = true
+basepython = python2.7
```
danlamanna: See my comment below.

tox.ini
```diff
@@ -0,0 +1,88 @@
...
+  --cov-config=.coveragerc.pytest \
+  # Surpress printing coverage, defer this to testenv:coverage-report
+  --cov-report= \
+  {posargs}
+
+# This report is used to merge coverage reports from testenv and from
+# the legacy tests run by ctest. When just running 'tox' it will print
+# the pytest coverage report. When running ctest && tox it will print
+# the combine coverage.
+[testenv:coverage-report]
```
danlamanna: This environment can't be run twice in a row (coverage combine fails after it's run once).

.coveragerc.pytest
kotfic: Coverage is run with three different config files,
+ pytest      -  .coveragerc.pytest
+ ctest        -  {build}/girder_worker/core/tests/girder_worker.coveragerc
+ combine  -  no config file (i.e. defaults)

the pytest coverage ignores all files in girder_worker/core/ and girder_worker/plugins/. If you try and combine the pytest coverage (which ends up in .coverage.pytest)  with the ctest coverage (which ends up in .coverage.legacy) with the .coveragerc.pytest config file,  then it ignores everything the ctest coverage created.  

Eventually .coveragerc.pytest will become .coveragerc,  but not until core and plugins are transitioned out of girder_worker.

girder_worker/core/tests/CMakeLists.txt
```diff
@@ -10,23 +10,23 @@ if(PYTHON_COVERAGE)
...
     NAME py_coverage_combine
     WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
-    COMMAND "${PYTHON_COVERAGE_EXECUTABLE}" combine
+    COMMAND "${PYTHON_COVERAGE_EXECUTABLE}" combine "--rcfile=${py_coverage_rc}"
   )
   add_test(
     NAME py_coverage
     WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
-    COMMAND "${PYTHON_COVERAGE_EXECUTABLE}" report --fail-under=${COVERAGE_MINIMUM_PASS}
+    COMMAND "${PYTHON_COVERAGE_EXECUTABLE}" report "--rcfile=${py_coverage_rc}" --fail-under=${COVERAGE_MINIMUM_PASS}
```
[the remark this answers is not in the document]
kotfic: fail-under could be moved,  but the rcfile now has to point to the ```{build}/girder_worker/core/tests/girder_worker.coveragerc``` file otherwise it tries to report from the 'default' data file (.coverage)  which no longer exists.

tox.ini
kotfic: is there a way to depend on other tests being run first with tox?

tox.ini
kotfic: Yes,  setting this basepython is a temporary measure.  it will be removed in a subsequent PR that makes the modern girder_worker code python2/3 compliant

PR #201, opened 2017-10-05
Add docker_run task

cjh1:

PR #226, opened 2017-12-18
Fallback to a regular container if nvidia is unavailable.

manthey:
When running an nvidia container, if we can't connect to nvidia, run as a regular container.

Ideally, it would be good if we knew if nvidia was *required* or only *preferred*.  When it is just preferred, it is better to run without nvidia support then to not run at all.

Note: PR #201 undid this change which had been added in PR #206.

### DIFF
--- a/girder_worker/docker/tasks/__init__.py
+++ b/girder_worker/docker/tasks/__init__.py
@@ -55,6 +55,14 @@ def _run_container(image, container_args,  **kwargs):
                 % (image, container_args, kwargs))
     try:
         return client.containers.run(image, container_args, **kwargs)
+    except nvidia.NvidiaConnectionError:
+        try:
+            logger.info('Running nvidia container without nvidia support: image: %s' % image)
+            client = docker.from_env(version='auto')
+            return client.containers.run(image, container_args, **kwargs)
+        except DockerException as dex:
+            logger.error(dex)
+            raise
     except DockerException as dex:
         logger.error(dex)
         raise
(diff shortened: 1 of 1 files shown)


kotfic:
@cjh1 was there a reason this was dropped?

girder_worker/docker/tasks/__init__.py
```diff
@@ -55,9 +55,17 @@ def _run_container(image, container_args,  **kwargs):
                 % (image, container_args, kwargs))
     try:
         return client.containers.run(image, container_args, **kwargs)
+    except nvidia.NvidiaConnectionError:
+        try:
+            logger.info('Running nvidia container without nvidia support: image: %s' % image)
+            client = docker.from_env(version='auto')
+            return client.containers.run(image, container_args, **kwargs)
+        except DockerException as dex:
```
zachmullen: Shouldn't this just be:

```
except DockerException:
    logger.exception('Exception when running without nvidia support.')
```

?

manthey: When we aren't using an nvidia container or are using one when nvidia support is available, we log a `DockerException` via `logger.error(dex)`.  I wasn't trying to change that behavior.  

In both cases, calling `logger.exception` probably makes more sense; I can change that as part of this PR.

manthey: Or, perhaps we should make that change in a different PR?

cjh1:
@kotfic Just slipped through the cracks, eventually it would be nice to have a test for this, so I don't break it again :smile:

girder_worker/docker/tasks/__init__.py
zachmullen: Let's go ahead and do it on this one if that's OK.

PR #232, opened 2018-01-11
Python 3 support

zachmullen:
In addition to making the modern g_w code work in python 3, this also makes several changes to our CI infrastructure and policies.

* This cleanly separates legacy code testing and modern testing. Legacy code testing is performed solely on travis, and modern code testing is performed solely on CircleCI. This means CMake/CTest are not run at all on CircleCI.
* The requirements.txt file (the lock file that we use for CI) was recompiled to exclude the legacy plugins. That file now represents the lock file for the modern framework only.
* Coverage reports are only uploaded to codecov (and status reports are only issued) from the modern codebase, not the legacy codebase. These coverage reports are only what was covered by unit tests, not integration tests.

I have decided that the task of adding python 3 integration tests to our CI process should wait for a future topic to make both topics easier to review. That future task should be straightforward after these changes.


zachmullen:
I can't figure out why the integration test is failing, anyone have an idea?

zachmullen:
@girder/core @girder/developers PTAL, this is now ready for review. Once this is merged I will mark the circleci status lines as required for future PRs.

kotfic:
This will need a follow on PR that adds python3 support to the integration tests.

kotfic:
I have been playing around with this locally trying to get the integration tests to pass.  It is looking like there are several string/byte related errors w.r.t. to the JobManager class and the docker infrastructure

kotfic:
I have opened https://github.com/girder/girder_worker/pull/235 against this branch fixing some issues i've encountered running the integration tests.

zachmullen:
I can't find anywhere in our modern CI that we actually disable `core_tasks`... maybe loading the tasks is python 3 compatible, although running them is not.

kotfic:
loading is actually not possible.  The modern unit tests don't actually load any tasks

zachmullen:
Do we disable them in the integration tests? I can't figure out where that's happening if so

kotfic:
The integration tests do not run in CI,  i had to set ```core_tasks = false``` to get them to run.  I am putting together a small PR now that will automatically disable core tasks if six.PY3 is True.

PR #233, opened 2018-01-12
Set the test infrastructure socket host to 0.0.0.0

kotfic:
Without this girder binds to 127.0.0.1 which does not forward through
the container and makes girder inaccessible to girder_worker (and the
host).

### DIFF
--- a/tests/integration/scripts/setup.yml
+++ b/tests/integration/scripts/setup.yml
@@ -19,6 +19,14 @@
         editable: true
 
 
+    # Note that this will force a server restart
+    - name: Set the server.socket_host to 0.0.0.0
+      lineinfile:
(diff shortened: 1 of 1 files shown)


kotfic:
See: https://github.com/girder/girder/pull/2565

danlamanna:
> @danlamanna @girder/developers this fixes things downstream in girder_worker... is this something we need to also fix in Girder's docker-compose setup?

Yes. I think the best way to support this might be to extend `girder-server` to accept a host argument, currently it only supports passing the port.

danlamanna:
Related upstream PR: https://github.com/girder/girder/pull/2570

zachmullen:
I'd like to revert this once https://github.com/girder/girder/pull/2570 is pushed to all the relevant places.

PR #238, opened 2018-01-15
Disable core tasks in py3

kotfic:
@jbeezley @zachmullen PTAL

### DIFF
--- a/girder_worker/__main__.py
+++ b/girder_worker/__main__.py
@@ -1,3 +1,4 @@
+import six
 from six.moves.configparser import NoSectionError, NoOptionError
 
 from . import config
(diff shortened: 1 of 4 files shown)


girder_worker/__main__.py
```diff
@@ -6,11 +7,14 @@
...
-    except (NoSectionError, NoOptionError):
-        include_core_tasks = True
+    if six.PY2:
+        try:
+            include_core_tasks = config.getboolean(
+                'girder_worker', 'core_tasks')
+        except (NoSectionError, NoOptionError):
+            include_core_tasks = True
+    else:
+        include_core_tasks = False
```
jbeezley: Would it make sense to print a warning if core tasks are enabled in the config in PY3?

tests/task_plugin_test.py
```diff
@@ -1,138 +1,145 @@
...
+import mock
+import pytest
+import six
+
+
+def setup_function(func):
+    if hasattr(func, 'namespace'):
+        namespace = func.namespace.args[0]
+        func.original = entrypoint.NAMESPACE
+        entrypoint.NAMESPACE = namespace
```
jbeezley: This doesn't appear to be used anywhere.

jbeezley: Oh, this is probably a pytest thing...  Ignore me.

kotfic: This is run before each test automatically by pytest.  It checks for the namespace mark on the test function and then sets the namespace to that.  ```teardown_function``` cleans it up like a context manager

girder_worker/__main__.py
kotfic: hopefully all of the logic related to ```core_tasks``` is going away in the next six to eight weeks.  For the most part i doubt people will be running girder_worker in python3 anyways.  This is mostly to prevent core tasks from being imported in girder's python3 test suite

PR #239, opened 2018-01-16
Add Dockerfile.py3 for building python3 based girder_worker image

kotfic:


Dockerfile.py3
```diff
@@ -0,0 +1,48 @@
...
+RUN useradd -D --shell=/bin/bash && useradd -m worker
+
+# RUN /usr/local/bin/girder-worker-config set girder_worker tmp_root /tmp
+RUN chown -R worker:worker /girder_worker
+
+USER worker
+
+RUN cp /girder_worker/girder_worker/worker.dist.cfg /girder_worker/girder_worker/worker.local.cfg && \
+    sed -i girder_worker/worker.local.cfg \
+    -e '/^broker/ s/guest@localhost/%(RABBITMQ_USER)s:%(RABBITMQ_PASS)s@%(RABBITMQ_HOST)s/'
```
zachmullen: Is there a way we could pass this as a command line argument somewhere instead of having to use advanced hackery to edit the config file?

kotfic: Yeah,  one of the things i'd like to do is rework the entire config mechanism in girder worker,  but for now this is just focused on replicating the python2 steps in a python3 Dockerfile.

zachmullen: If there isn't a runtime invocation command to use, there's always the `girder-worker-config set` command that could be used here instead of `sed`.

PR #240, opened 2018-01-16
Append py27 and py36 coverage

kotfic:


zachmullen:
This is for local testing, I'm assuming?

kotfic:
Addition of the coverage-erase target is for local testing yes. The ```--cov-append``` flag makes sure our coverage reflects coverage from both PY2 and PY3 environments locally and on circle.

zachmullen:
There is divergence in our use of tox between local and CI environments, however that may be an expected thing.

kotfic:
coverage-erase in CI isn't run and the ```--cov-append``` flag in testenv basically does nothing in CI.

PR #241, opened 2018-01-16
Use girder cli host flag

zachmullen:
This should fail until the new girder/girder build on docker hub completes.

ping @kotfic @danlamanna


kotfic:
happy to approve once integration_tests pass

zachmullen:
good to go

PR #244, opened 2018-01-17
Transition document

kotfic:
Adds documentation specifically targeted at transitioning legacy girder worker tasks

See: [TRANSITIONING.md](https://github.com/girder/girder_worker/blob/transition-document/TRANSITION.md) for a rendered version of the docs


kotfic:
@brianhelba @zachmullen @mgrauer @danlamanna  I have not written the section on transitioning docker mode legacy tasks,  but once I do I am hoping you'll be able to use this document to transition some tasks and provide feedback - merging only once the pilot tasks are transitioned and the document is looking good.

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+from girder_worker.app import app
+from girder_worker.utils import girder_job
+
+@girder_job(title='Add a and b')
+@app.task(bind=True)
+def my_task(self):
+	return a + b
+```
+
+Then the legacy code above may be rewritten as follows:
```
jeffbaumes: The "legacy code above" is actually two code blocks above which is a bit confusing.

I suggest moving the task definition to the previous section as an example of what an implemented task function might look like (i.e. how to change the template), then here refer to that same task example.

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+
+return async_result.job
+```
+
+For more details about manipulating Job attributes,  please read [The Celery Method](https://github.com/girder/girder_worker/blob/transition-document/HACKING.md#the-celery-method) section of the HACKING document.
+
+
+### Transformations
+
+The following steps are sufficient for converting legacy Girder Worker tasks into modern Celery based tasks. If however your task requires access to information inside Girder,  then it is recommended that you include "Transform" classes with your package.
```
jeffbaumes: "following" -> "previous"

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+def my_task(self):
+	return a + b
+```
+
+Then the legacy code above may be rewritten as follows:
+
+```python
+from my_package.tasks import example_task
+
+async_result = my_task.delay(1, 2)
```
danlamanna: `example_task`

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+
+## "Docker" Mode Legacy Tasks
+
+
+
+
+
+
+## "Other" Mode Tasks
+Girder Worker no longer explicitly supports tasks written in other modes (e.g. R, scala, etc). Custom tasks could be written to support these modes based on the legacy executor.py file that implements that mode. For more information please contact the D&A team.
```
jeffbaumes: For a public document, "contact the D&A team" should probably instead be a link to the Girder Gitter channel and/or Discourse instance.

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+
+3. What about Girder I/O? How do I push and pull data?
+  Please use Transforms to manage this process. The base Transform class is implemented in girder\_worker\_utils.transform.Transform. Currently there are several example implementations of transforms in girder\_worker\_utils.transforms.girder_io. We are hoping that as more projects implement common useful Transforms that these will be upstreamed into girder\_worker\_utils.
+
+## "Docker" Mode Legacy Tasks
+
+
+
+
+
```
jeffbaumes: I think there is another section needed to deal with item_tasks conversions once available, which I believe will need type hinting to be in the installed task package in order to generate an adequate UI. Currently the type hinting in your example legacy code (e.g. inputs specified as "integer") is left without a replacement mechanism when ported over. This would be a layer which we should make clear is not required, but is explicitly to provide metadata for UI gen tools.

TRANSITION.md
jeffbaumes: I wonder if most of this section should be made the official "Transforms" documentation in another location, and that this document should just point to it. When porting, I expect the "80%" case of transforms will already be available in the utils package, so all people don't need to know the internal mechanics of how they are implemented.

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+async_result = my_task.delay(
+    1, 2, girder_job_title="A Custom Job Title")
+
+return async_result.job
+```
+
+For more details about manipulating Job attributes,  please read [The Celery Method](https://github.com/girder/girder_worker/blob/transition-document/HACKING.md#the-celery-method) section of the HACKING document.
+
+
+### Transformations
```
jeffbaumes: Rename to "Transforms" for consistency?

TRANSITION.md
```diff
@@ -0,0 +1,363 @@
...
+```
+
+This provides a generic mechanism to handle side effects on the results of a task function at the time the task is launched.
+
+
+### FAQ
+
+1. [What if my task has complex install dependencies?](HACKING.md#what-if-my-task-has-complex-install-dependencies)
+
+2. [How do I write a custom task from scratch (e.g. without the cookiecutter)](HACKING.md#what-if-my-task-has-complex-install-dependencies). A related question: What is the cookiecutter actually doing?
```
jeffbaumes: Is this supposed to link to the same section in HACKING.md as item 1?

matthewma7:
I don't know if this is the place to leave a feedback. 
I think it is `girder_result_hooks` in the code instead of `girder_results_hook`.

TRANSITION.md
brianhelba: @jeffbaumes I think this feedback is beyond the scope of this PR. Can you open it as an issue?

TRANSITION.md
brianhelba: @jeffbaumes FYI, I wasn't able to directly resolve this via my edits, but I agree there's room for further improvement. If you're able to improve the flow of this section, please do so in another PR.

brianhelba:
@kotfic I've made some major edits and addressed all the feedback that I could. PTAL now.

I'd like to go ahead and merge this now, if you don't have significant objections. We can always refine it further via subsequent PRs.

brianhelba:
@jeffbaumes FYI, other than where I explicitly replied and noted otherwise, I was able to address all of your other comments / feedback with my edits.

mgrauer:
🌵 🦋

mgrauer:
I should probably elaborate for those without poetic sensibilities...

Awesome work @kotfic and thanks to @brianhelba and @jeffbaumes for detailed QA.

PR #246, opened 2018-01-18
Filename option for GirderFileIdToVolume

zachmullen:
@kotfic @cjh1 PTAL

### DIFF
--- a/girder_worker/docker/transforms/girder.py
+++ b/girder_worker/docker/transforms/girder.py
@@ -47,22 +47,23 @@ class GirderFileIdToStream(GirderClientTransform):
 
 
 class GirderFileIdToVolume(GirderClientTransform):
-    def __init__(self, _id, volume=TemporaryVolume.default, **kwargs):
+    def __init__(self, _id, volume=TemporaryVolume.default, filename=None, **kwargs):
         super(GirderFileIdToVolume, self).__init__(**kwargs)
         self._file_id = _id
         self._volume = volume
+        self._filename = filename
         self._file_path = None
 
     def transform(self, **kwargs):
         self._volume.transform(**kwargs)
         dir = self._volume.host_path
-        self._file_path = os.path.join(
-            dir, self._file_id)
+        filename = self._filename or self._file_id
+        self._file_path = os.path.join(dir, filename)
 
         self.gc.downloadFile(self._file_id, self._file_path)
 
         # Return the path inside the container
-        return os.path.join(self._volume.container_path, self._file_id)
+        return os.path.join(self._volume.container_path, filename)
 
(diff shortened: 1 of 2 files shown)


girder_worker/docker/transforms/girder.py
```diff
@@ -47,22 +47,23 @@ def transform(self, **kwargs):
 
 
 class GirderFileIdToVolume(GirderClientTransform):
-    def __init__(self, _id, volume=TemporaryVolume.default, **kwargs):
+    def __init__(self, _id, volume=TemporaryVolume.default, filename=None, **kwargs):
```
cjh1: Are we happy with the user having to pass the filename in when we could extract if from Girder?

zachmullen: I considered that as another option. Ideally both would be possible. Maybe another option `use_girder_filename`?

cjh1: I was thinking that if filename was ```None``` we could get it from Girder?

zachmullen: I'd be OK with that, though it breaks the existing behavior.

cjh1: I guess one issue with that approach is filename conflicts, where as the default of using the uuid is going to be pretty safe from that point of view.

zachmullen: Right.

cjh1: I guess we could get the best of both worlds with <root>/<file_id>/<filename>?

zachmullen: I like that plan, it also mitigates the risk of any output paths conflicting with input paths.

I'll do that extra directory as the default case, but if a filename is passed explicitly, we'll leave out the UUID. That will allow clients max control, which may be useful in odd cases like where two files need to be in the same directory (e.g. mha/mhd read side effects).

cjh1: Sounds good.

zachmullen:
Due to the requested changes, there are several conflicts with #247. Since they are so closely related, I'm going to just combine the two PRs into one.

cjh1:
@zachmullen This looks good, the tests need to be updated to reflect the path changes.

zachmullen:
Because we are now actually using `mkdir`, the errors are mostly resulting from trying to `mkdir` in a bogus directory. My plan was to just mock out `os.mkdir` in the tests, do you have a better method?

cjh1:
That will work or prefix the directories with /tmp.

zachmullen:
The tests have been fixed, this is ready for final review

Issue #249, opened 2018-01-18
docker_run job succeeds even though container exits with non-zero


zachmullen:
Related to this, if the docker container exits with non-zero, the output hooks still attempt to run, which probably shouldn't happen.

zachmullen:
@cjh1 after looking at the docs for docker-py, we don't get an exception for free because we use `detach=True` on the container. `container.wait()` will block until the container stops and return its return code, however I'm not sure where to put that logic in our docker task so that it's safe. Any thoughts?

cjh1:
@zachmullen We don't want do ```container.wait()``` while the container is running as we need the thread to run the event loop. However, after the container has exited we can get the exit code using the wait call and raise an exception if its non zero. I am thinking around [here](https://github.com/girder/girder_worker/blob/master/girder_worker/docker/tasks/__init__.py#L151). If you don't get to this I am happy to take a look next week.

zachmullen:
Ok, great. I was mainly wondering whether it should go inside _run_select_loop or after it.

PR #250, opened 2018-01-18
Raise exception if docker exits with non-zero code

zachmullen:
Here's a candidate solution to #249 . I'm not sure how it interacts with job cancellation or other odd circumstances surrounding docker container execution, but it works in the case of the container returning normally with non-zero.

### DIFF
--- a/tests/integration/test_docker.py
+++ b/tests/integration/test_docker.py
@@ -5,6 +5,28 @@ import json
 from girder_worker.utils import JobStatus
 import pytest
 
+FIXTURE_DIR = os.path.join('..', os.path.dirname(__file__), 'fixtures')
+
+
+def _assert_job_statuses(job):
+    assert [ts['status'] for ts in job['timestamps']] == [JobStatus.RUNNING, JobStatus.SUCCESS]
+
+
+def _assert_job_contents(r, session, test_file, remove_newline=True):
+    assert r.status_code == 200, r.content
+
+    with session.wait_for_success(r.json()['_id']) as job:
+        _assert_job_statuses(job)
+
+        # Remove escaped chars
+        log = ''.join(str(l) for l in job['log'])
+        # Remove trailing \n added by test script
+        if remove_newline:
+            log = log[:-1]
+
+        with open(test_file) as fp:
+            assert log == fp.read()
+
 
 @pytest.mark.docker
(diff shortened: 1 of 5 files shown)


zachmullen:
I can probably come up with an integration test for this, but I'm unsure how to unit test it -- looks like everything in `test_docker_tasks.py` is currently stubbed.

cjh1:
@zachmullen I think an integration test is the best approach here.

zachmullen:
@cjh1 integration test added, PTAL

cjh1:
@zachmullen This is looking good, thanks for the test de-duplication :-) It occurs to me that there may be another use case that we should probably think about. The developer may actually want to get hold of the docker exit code. I am thinking of the case where you have a couple of task chained together. If the docker running fails then the return code might be used by the next task to determine what to do next. In this case you just want to return the exit code from the task rather than raise an exception. Now in this case would you mark the task's state as error? I could argue either way. We could control this behavior by passing in another parameter, something like ```raise_on_error```, defaulting to True. Thoughts?

zachmullen:
I do think having a feature to control this would be good. Is that something you could add as a follow-up to this PR?

cjh1:
@zachmullen Absolutely, I wasn't suggesting adding it here. I will raise an issue.

cjh1:
I have raised #251

tests/integration/test_docker.py
```diff
@@ -112,51 +102,15 @@ def test_docker_run_file_upload_to_item(session, girder_client, test_item):
 
 
 @pytest.mark.docker
-def test_docker_run_girder_file_to_named_pipe_on_temp_vol(session, test_file, test_file_in_girder):
-    """
-    This is a simplified version of test_docker_run_girder_file_to_named_pipe
-    it uses the TemporaryVolume, rather than having to setup the volumes
-    'manually', this is the approach we should encourage.
```
cjh1: Was there a reason this doc string was removed?

zachmullen: Kind of an unintended consequence of collapsing the two tests into one via `parametrize`... I don't fully understand the comment, do you think it's important to have it here inline?

cjh1: If you don't understand it then its best left removed :smile:

PR #252, opened 2018-01-19
Py3 integration tests

kotfic:
Add testing to circle for running integration tests in python2 and python3


zachmullen:
I think the python3 integration tests are actually running in python 2 on CircleCI:

```
#!/bin/bash -eo pipefail
cd tests/integration && make test3
pytest -v -x -n4 -k "not traditional"
============================= test session starts ==============================
platform linux2 -- Python 2.7.12, pytest-3.1.0, py-1.5.2, pluggy-0.4.0 -- /opt/circleci/.pyenv/versions/2.7.12/bin/python2.7
```

See https://circleci.com/gh/girder/girder_worker/784

kotfic:
Looks like ```test_docker_run_file_upload_to_item[admin-admin]``` is failing with ```OSError: [Errno 9] Bad file descriptor``` I am not seeing the error locally when running the integration tests in python 3. I'm am looking into it

kotfic:
Actually,  this is looking dicey enough that i am going to try and address it monday with @cjh1 -  As a note to Monday kotfic - this behavior started when switching pytest from running in 2.7.13  to 3.6.3 -  You would like to think that the tests are just using requests to hit endpoints and so it shouldn't matter -  but actually some of the docker tests are setting up temporary volumes etc which may be the root of the problem.

cjh1:
#258 Should fix up the errors we are seeing here.

PR #257, opened 2018-01-22
Add some model string representations

### LINKED ISSUE #243: Implement __str__ for docker mode classes
In particular, I now see the following kwargs in new-style docker_run jobs in Girder:

```
{
  "container_args": [
    "--phi-samples",
    "12",
    "--theta-samples",
    "3",
    "--width",
    "256",
    "--height",
    "256",
    "--preset",
    "CT-AAA",
    "<girder_worker.docker.transforms.girder.GirderFileIdToVolume object at 0x10eb6eb90>",
    "<girder_worker.docker.transforms.VolumePath object at 0x10eb6e750>"
  ]
}
```

This makes it difficult to debug issues related to the mounted volumes. An informative `__str__` (or `__repr__`, whichever one is getting used here) would be quite useful.

zachmullen:
Fixes #243 

This addresses the couple of classes I am using in my current use case.

### DIFF
--- a/girder_worker/docker/transforms/__init__.py
+++ b/girder_worker/docker/transforms/__init__.py
@@ -318,6 +318,9 @@ class VolumePath(Transform):
         else:
             return os.path.join(self._volume.container_path, self.filename)
 
+    def _repr_model_(self):
+        return '<%s.%s: "%s">' % (self.__module__, self.__class__.__name__, self.filename)
+
 
 class Connect(Transform):
     """
--- a/girder_worker/docker/transforms/girder.py
+++ b/girder_worker/docker/transforms/girder.py
@@ -85,6 +85,15 @@ class GirderFileIdToVolume(GirderClientTransform):
         if self._file_path is not None:
             shutil.rmtree(self._file_path, ignore_errors=True)
 
+    def _repr_model_(self):
+        if self._filename:
+            template = u'<{module}.{cls}: File ID={id} -> "{fname}">'
+        else:
+            template = u'<{module}.{cls}: File ID={id}>'
+        return template.format(
+            module=self.__module__, cls=self.__class__.__name__, id=self._file_id,
+            fname=self._filename)
+
 
 class GirderUploadVolumePathToItem(GirderUploadToItem):
     def __init__(self, volumepath, item_id,  delete_file=False, **kwargs):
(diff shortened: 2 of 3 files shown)


girder_worker/docker/transforms/__init__.py
```diff
@@ -318,6 +318,9 @@ def transform(self, *pargs, **kwargs):
         else:
             return os.path.join(self._volume.container_path, self.filename)
 
+    def _repr_model_(self):
+        return '<%s.%s: "%s">' % (self.__module__, self.__class__.__name__, self.filename)
```
jeffbaumes: Seems odd that one `_repr_model_` uses the pattern `'str' % (args,)` while the other uses `u'str'.format()`. Note the different use of `u` prefix and `format`/`%`. Are we moving toward one method of formatting strings or are there reasons for the difference?

kotfic: I agree with Jeff that these should be consistent.  My preference would be for ```format``` but that is just because of personal habit,  not because of a reasoned argument.

zachmullen: There are several ways to concatenate strings in python: `+`, `%`, `format`, `join`... Of these, `format` is the only one that does not automatically upgrade the output type to text type if one of the inputs happens to be text type. For that reason, we avoid using it in Girder and prefer `%` for formatting even though it lacks power -- if one forgets to prepend the `u` type specifier to the template string, it can cause exceptions to occur that won't be caught in testing unless you test with strings outside the range of ASCII.

This felt like a good time to break that rule because of the variable template. However, consistency is desirable. I'm fine with having them both use one way or the other, but if `format` is the way we want to go everywhere, we need to be disciplined about enforcing the `u` before the template string. I would go as far as wanting a lint rule for it if that's possible.

kotfic: An alternative would be to ```from __future__ import unicode_literals``` which would force all strings to be unicode literals unless they were specifically declared with b''. 

Currently the majority of the string formatting is done with ```%```  so we should probably stick with that for consistencies sake. 

As a side note - i'm of the opinion that automatically upgrading to text types is a _bug_ not a feature.  It is exactly the kind of behavior that allows Python 2 applications to be lazy about their bytes/text API, making converting to Python 3 difficult.

zachmullen: > automatically upgrading to text types is a bug not a feature. It is exactly the kind of behavior that allows Python 2 applications to be lazy about their bytes/text API, making converting to Python 3 difficult.

I kind of agree with this, but not fully. I would fully agree if `unicode_literals` was the default behavior, but since it's not, I think this behavior creates too many pitfalls that won't be discovered until runtime.

jeffbaumes:
Agreed @kotfic sorry for the noise. Perhaps Python conventions that aren't lintable should be documented somewhere, but at some point the minutia get too minute to bother with.

PR #260, opened 2018-01-24
Refactoring JobManager to use Tee objects for stdout/err redirection

kotfic:
Depends on https://github.com/girder/girder_worker_utils/pull/15 - this is still evolving rapidly as changes happen to the GWU PR


tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
+import sys
+from girder_worker.utils import TeeStdOutCustomWrite, TeeStdErrCustomWrite, JobManager
+
+
+def test_TeeStdOutCustomeWrite(capfd):
+    _nonlocal = {'data': ''}
```
zachmullen: The python convention for avoiding reserved words is to use a suffix _, so `nonlocal_` rather than `_nonlocal`.

(I don't actually care, just feeling pedantic.)

tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def _append_to_data(message, **kwargs):
+        _nonlocal['data'] += message
+
+    with TeeStdOutCustomWrite(_append_to_data):
+        sys.stdout.write('Test String')
+        sys.stdout.flush()
+
+    assert _nonlocal['data'] == 'Test String'
```
zachmullen: Can we also assert that the "real" stdout was written to by using the `capsys` fixture?

kotfic: sure,  though actually we should probably check the ```capfd``` fixture,  as i'm not sure how our hijacking and pytest's hijacking will mix.

kotfic:
@zachmullen this is ready for another look. Thanks!

requirements3.txt
```diff
@@ -0,0 +1,88 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --output-file requirements3.txt requirements.in requirements-dev.in
```
zachmullen: I know this is an auto-generated comment, but the inputs listed are the same as for `requirements2.txt`, so it would be unclear to someone reading this how they might update this file for both python 2 and 3. It's probably worth mentioning how to accomplish this in the docs somewhere.

tests/integration/scripts/setup.yml
```diff
@@ -13,6 +13,7 @@
         name: /girder_worker
         editable: true
 
+
```
zachmullen: Any reason for this? Seems to be inconsistency in white space usage in this file.

tox.ini
```diff
@@ -9,8 +9,9 @@ basepython = python2.7
 deps = coverage
 commands = coverage erase --rcfile=.coveragerc.pytest
 
+
+
```
zachmullen: Do we need 3 blank lines here?

requirements3.txt
kotfic: After speaking with @jbeezley he would like to transition girder worker to use [pipfile](https://github.com/pypa/pipfile) so i'm not sure how long this will be around - see also https://github.com/girder/girder_worker/issues/265

zachmullen: Ok, we'll punt on this.

Issue #263, opened 2018-01-26
docker_run does not remove container after task is done


zachmullen:

cjh1:
Even with remove_container=True?

zachmullen:
Well this is a bit embarrassing

zachmullen:
Should we make that default to True?

PR #264, opened 2018-01-26
Remove container by default in docker mode

zachmullen:
refs #263 

I think this used to be the default in legacy mode, and I think it's still a sensible default -- I don't see a lot of utility in production in keeping the dead containers around, though if we add some logging of the container ID back to the job document it could be a very useful feature for debugging.

### DIFF
--- a/girder_worker/docker/tasks/__init__.py
+++ b/girder_worker/docker/tasks/__init__.py
@@ -281,7 +281,7 @@ class DockerTask(Task):
 
 
 def _docker_run(task, image, pull_image=True, entrypoint=None, container_args=None,
-                volumes={}, remove_container=False, stream_connectors=[], **kwargs):
+                volumes={}, remove_container=True, stream_connectors=[], **kwargs):
 
     if pull_image:
         logger.info('Pulling Docker image: %s', image)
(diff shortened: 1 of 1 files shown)


danlamanna:
Related to this area of the code, is there a benefit to having a separate cleanup step instead of passing  `remove=remove_container` to the `docker.run` command and letting docker handle it?

zachmullen:
Yes, because the container may have written to the mounted volume that still exists as a temp dir on the host that we still want to clean up. Not only that, we typically want to upload those files back to girder after the container exits, so we *need* them to outlive the container process.

The reason we have to clean them up using another docker image is because the docker engine writes those files with owner=root, and we prescribe not running the docker client as root.

cjh1:
@zachmullen Agreed for some reason I thoughtthe default was not to clean them up. As you say there is little point in the used containers piling up in production.

zachmullen:
@cjh1 [you are correct](https://github.com/girder/girder_worker/blob/master/girder_worker/plugins/docker/executor.py#L203) about the legacy behavior, though I feel like at some point in the past it did the removal by default.

Thanks for the review.

PR #268, opened 2018-01-30
Fixes #266. Allow use of default container_args param

### LINKED ISSUE #266: docker_run tasks fail with default arguments
When a docker_run task is run with its default argument (None) for `container_args`, it raises a TypeError, since it expects `container_args` to be an iterable.

zachmullen:
Also remove mutable default param values in docker_run.

ping @wphicks


cjh1:
@zachmullen Thanks for fixing that up :frowning:

PR #271, opened 2018-01-31
Restructure of app.py

kotfic:
Currently there is too much going on in app.py.  This PR attempts to make no changes except reorganizing the code to be more readable.

In particular it introduces the notion of a ```context```  which is a module in the context subpackage that implements a well defined API.  The goal is to handle the several places in the code where what we want to do is effected by whether or not we are currently in a girder rest request.

### DIFF
--- a/girder_worker/context/__init__.py
+++ b/girder_worker/context/__init__.py
@@ -3,6 +3,12 @@ from . import girder_context, nongirder_context
 
 def get_context():
     try:
+        # Note: If we can import these objects from the girder packages we
+        # assume our producer is in a girder REST request. This allows
+        # us to create the job model's directly. Otherwise there will be an
+        # ImportError and we can create the job via a REST request using
+        # the jobInfoSpec in headers.
+
         from girder.utility.model_importer import ModelImporter # noqa
         from girder.plugins.worker import utils # noqa
         from girder.api.rest import getCurrentUser # noqa
--- a/girder_worker/context/__init__.py
+++ b/girder_worker/context/__init__.py
@@ -1,0 +1,11 @@
+from . import girder_context, nongirder_context
+
+
+def get_context():
+    try:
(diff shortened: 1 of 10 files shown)


kotfic:
@dorukozturk is [this](https://github.com/girder/girder_worker/blob/app-refactor/girder_worker/app.py#L40-L45) more readable?   If so,  i'll convert the other before_task_publish steps (as well as the try/import exception in GirderAsyncResult)

kotfic:
Strange.. tests passed locally...

girder_worker/context/__init__.py
```diff
@@ -0,0 +1,11 @@
+from . import girder_context, nongirder_context
+
+
+def get_context():
+    try:
+        from girder.utility.model_importer import ModelImporter # noqa
+        from girder.plugins.worker import utils # noqa
+        from girder.api.rest import getCurrentUser # noqa
```
cjh1: Isn't line 6-7 enough to ensure we are in a Girder context with the worker plugin enable?

kotfic: probably,  actually i'd like to change this to check if there is an actual request object available.  Otherwise running in a virtualenv with girder causes a false positive on this test

danlamanna:
I think this is a big step forward in terms of making app.py more approachable.

I would suggest some consideration towards naming though, something like:
`handle_jobInfoSpec` -> `create_task_job`
`handle_girder_api_url` -> `get_girder_api_url`
`handle_girder_client_token` -> `attach_girder_client_token`

girder_worker/context/__init__.py
kotfic: See: https://github.com/girder/girder_worker/issues/274

mgrauer:
@kotfic Can you add a post-humous CHANGELOG entry for this?

Issue #270, opened 2018-01-31
Provide support for runtime creation of Docker-managed volumes


wphicks:
It should be possible to create a Docker-managed volume (not a bind-mount) and use it in the context of a girder_worker instance running inside Docker that launches Docker tasks.

kotfic:
This would resolve issues where the worker docker container needs to mount the host /"tmp" inside the container

zachmullen:
@kotfic that can be accomplished with bind mounts. @wphicks could you describe your use case in more detail so we know what sort of API we'll need to expose?

cjh1:
@kotfic Not sure I understand the issues with the worker docker containers needing to mount the host "/tmp" ? The user doesn't have to do this? It just provides them with the ability to have a temporary directory shared between the host and container, this is a nice way to get results out of a container.

kotfic:
So this is applicable in situations where docker is running alongside docker (e.g  girder worker running in a container and a task running in a container).  Currently the TemporaryNamedVolume container tries to mount  the worker container's tmp directory into the task container's tmp directory.  Because the worker is talking to  the docker engine via a mounted socket file,  it is **actually** talking to the docker engine running on the host machine.  That means when the worker container makes the request to mount ```/tmp/whatever/foo.jpg``` inside the task container,  it is passed to the docker engine running on the host,  and the docker engine ***mounts the host's /tmp/whatever/foo.jpg directory*** inside the task container.  This requires that the host /tmp directory be mounted in the worker container's /tmp directory so that the host, the worker and the task all share the same /tmp directory. 

mounting the host /tmp directory into the worker /tmp directory must be performed *at deploy time* (for instance [here](https://github.com/girder/girder_worker/blob/master/tests/integration/docker-compose.yml#L33-L34).

Alternately,  if we used dockerpy  to create a docker manged volume,  that volume would exist on the host at /var/run/docker/....  and could be mounted *at run time*  inside the worker container  and the task container.

zachmullen:
Ah, understood, thanks for clarifying.

zachmullen:
Is this something we may need to break the current volume-related API for?

cjh1:
@kotfic that makes sense now.

kotfic:
[...]

PR #273, opened 2018-02-01
Rename Volume => BindMountVolume

cjh1:
To be more explicit about the sort of volumes we are dealing with and clear the way for supporting other volume types in the future.

### DIFF
--- a/docs/docker_run.rst
+++ b/docs/docker_run.rst
@@ -12,10 +12,10 @@ The ``docker_run`` task exposes a ``container_args`` parameter which can be used
 to pass arguments to the container entrypoint.
 
 
-Volumes
--------
+BindMountVolumes
+----------------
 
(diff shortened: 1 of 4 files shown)


cjh1:
Related to #270

zachmullen:
Technically a breaking change, though it doesn't break my use case (I just use `VolumePath`). I'm fine with it, but I'd like to know what the eventual API will look like -- will `BindMountVolume` and the real `Volume` share a base class, or be two totally different things?

cjh1:
@zachmullen  It is, but as we haven't reached 1.0 I figure it was worth doing now. The eventual API would probably have a Volume base class with sub classes of BindMountVolume and ManagedVolume or something similar.

zachmullen:
Ok, thanks.

Issue #274, opened 2018-02-01
Catching girder ImportException is not sufficient for determining signal context


kotfic:
If your worker is running in a virtualenv where girder is importable,  then you can find yourself in a situation where you should be working from a nongirder context but end up executing code in the girder context.

kotfic:
@wphicks this was the same issue you were seeing.

PR #275, opened 2018-02-07
Use cherrypy.request.app to determine producer/consumer context

kotfic:
Should fix https://github.com/girder/girder_worker/issues/274

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ Added Features
 Bug fixes
 ---------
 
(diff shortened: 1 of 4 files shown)


kotfic:
@zachmullen i'm not sure i follow,  do you have an example?

zachmullen:
I'm not wanting to hold this PR up, I think it's fine to merge, but it would be nice if we could do something like 

```
from girder import areWeInsideAGirderRequest


if areWeInsideAGirderRequest():
   ...
```

Rather than inferring it from something like cherrypy values being set.

PR #277, opened 2018-02-07
Continuous Delivery

kotfic:
Add continuous delivery similar to [girder_worker_utils](https://github.com/girder/girder_worker_utils/pull/18)


kotfic:
@danlamanna @brianhelba PTAL

tests/integration/Makefile
```diff
@@ -15,7 +15,9 @@ run3:
 	docker-compose -f docker-compose.yml -f docker-compose.py3.yml up -d; \
 	cd scripts/ && ansible-playbook -i inventory -e ansible_python_interpreter=python3 setup.yml
 
-test:
+test: test2
+
+test2:
```
danlamanna: What's the purpose of `test2`?

kotfic: makes an interface that is consistent (e.g.  ```make test2``` and ```make test3```). This will go away once the legacy code is removed.  For now though we need to provide a ```-k "not traditional"``` flag to pytest in python3.  Introducing ```test2``` allows us to use a single run command in the circle yaml:

```
run-tests: &run-tests
   run:
     name: Run integration tests
     command: |
       cd tests/integration && make test${RUN_ENV}
```

Where ```RUN_ENV``` is defined as an environment variable on the workflow

PR #278, opened 2018-02-08
Move task discovery out of __main__ into app initialization

kotfic:
While this is a small code change it's actually potentially a pretty big functionality change.

What this PR enables,  is the ability to use celery "as intended"  on the command line.  While ```girder-worker -l info```  still works,  this also supports running girder worker as ```celery worker -A girder-worker.app -l info ...```.  This is important because celery offers a number of other commands we might be interested in supporting/using in production environments including flower,  events,  inspect, control, among others. 

The primary thing preventing us from using those was task discovery (via entrypoints)  happening in the ```main()``` function  and not as apart of the import side-effects of the app module (which is what implicitly celery expects). 

This has some positive side effects. It resolves a lurking issue on the producer side,  where ```girder_worker.app.app``` was not being initialized with all the installed tasks,  but instead had tasks added to it as your imported them into the running girder instance For example ```from my_packagbe.tasks import my_task```.  Because my_task is wrapped by ```@app.task(...)``` it was being added to the internal ```app._tasks``` mapping at import time. But what this means is that ```app._tasks``` only reflects the imported tasks across all plugins/handlers/etc **not** the tasks that are actually pip installed inside the environment.  Most of the time that difference doesn't functionally matter,  but it is still unpleasantly confusing.  This resolves that issue,  but at the risk of surfacing other problems on the producer that were hidden until now. 

@zachmullen could you take a look in the context of the AaaS stuff? @cjh1's input would also be very welcome!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Changes
 
 * Volume transforms renamed to BindMountVolume (`#273 <https://github.com/girder/girder_worker/pull/273>`_)
 * Refactor signals to introduce 'context' in which code is being executed (`#271 <https://github.com/girder/girder_worker/pull/271>`_)
+* Move entrypoint based task descovery earlier in application initialization (`#271 <https://github.com/girder/girder_worker/pull/278?`_)
 
 
 Deprecations
--- a/girder_worker/__main__.py
+++ b/girder_worker/__main__.py
@@ -1,25 +1,7 @@
-from six.moves.configparser import NoSectionError, NoOptionError
-
-from . import config
 from .app import app
-from .entrypoint import get_core_task_modules, get_plugin_task_modules
 
 
 def main():
-    try:
-        include_core_tasks = config.getboolean(
-            'girder_worker', 'core_tasks')
-    except (NoSectionError, NoOptionError):
-        include_core_tasks = True
-
-    if include_core_tasks:
-        app.conf.update({
-            'CELERY_IMPORTS': get_core_task_modules()
-        })
-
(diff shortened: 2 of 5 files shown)


zachmullen:
While I do support this feature, I'm generally opposed to introducing import-time side effects... is there a way to do this without them, or is that just what we are stuck with?

cjh1:
@kotfic @zachmullen You might be able to do this task discovery as part of a [worker bootstep](http://docs.celeryproject.org/en/latest/userguide/extending.html#command-line-programs).

cjh1:
Also :+1: for being able to just run celery ...

kotfic:
@zachmullen that is just sort of how celery works.  The expectation is that you define and setup/configure a global ```Celery``` object,  in our case that includes task discovery through the entrypoint mechanism.  This can sort of be chalked up to "celery is an application"  rather than a library and i've done my best to isolate all import-time processing to ```girder_worker.app```

kotfic:
@cjh1 that is a great suggestion,  I will look into it - did you say thought that there were issues with bootsteps?

cjh1:
@kotfic Depends on the parts of the worker you need access to. I had problems requiring ```celery.worker.components:Timer``` and related components. I think for this you will probably be OK as your not going to have to reach very deep into Celery.

kotfic:
@cjh1 turns out blueprints won't work in this case,  We want to make sure we've discovered all the tasks installed on the system on both the producer side and the consumer side,  blueprints only run consumer side when you start the worker process

PR #280, opened 2018-02-15
Don't break chords

### LINKED ISSUE #279: Don't break chords
Running GW tasks in a chord, like so:
```
(celery.group(...) | other_task).delay()
```

Raises the following exception:
```
[2018-02-13 10:44:36,570: WARNING/ForkPoolWorker-4] Task celery.chord_unlock[25073a0a-0d2f-4de6-8183-7e6108a3497c] reject requeue=False: 'NoneType' object has no attribute 'update'
  File "/home/dan/.virtualenvs/gw/lib/python2.7/site-packages/celery/app/trace.py", line 374, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/dan/projects/gw/girder_worker/girder_worker/task.py", line 151, in __call__
    results = super(Task, self).__call__(*_t_args, **_t_kwargs)
  File "/home/dan/.virtualenvs/gw/lib/python2.7/site-packages/celery/app/trace.py", line 629, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/dan/.virtualenvs/gw/lib/python2.7/site-packages/celery/app/builtins.py", line 75, in unlock_chord
    raise self.retry(countdown=interval, max_retries=max_retries)
  File "/home/dan/.virtualenvs/gw/lib/python2.7/site-packages/celery/app/task.py", line 687, in retry
    try:
Reject: (AttributeError("'NoneType' object has no attribute 'update'",), False)
```

This happens because our custom `apply_async` expects `options` to be a dict. However, when called via `chord_unlock` it's `None` because `chord_unlock` creates a signature and calls `apply_async` from there.

Perhaps Celery builtin tasks should be totally avoided by our code paths, we probably don't want to create jobs for various Celery housekeeping tasks, for example.

kotfic:
This should fix #279 @danlamanna PTAL

### DIFF
--- a/girder_worker/utils.py
+++ b/girder_worker/utils.py
@@ -10,8 +10,20 @@ from requests import HTTPError
 import six
 
 
+BUILTIN_CELERY_TASKS = [
+    'celery.accumulate'
+    'celery.backend_cleanup',
+    'celery.chain',
+    'celery.chord',
+    'celery.chord_unlock',
+    'celery.chunks',
+    'celery.group',
+    'celery.map',
+    'celery.starmap']
+
+
 def is_internal(task):
-    return task.startswith('celery')
+    return task in BUILTIN_CELERY_TASKS
 
 
 def _maybe_model_repr(obj):
--- a/girder_worker/utils.py
+++ b/girder_worker/utils.py
@@ -10,6 +10,10 @@ from requests import HTTPError
 import six
 
 
+def is_internal(task):
+    return task.startswith('celery')
+
+
 def _maybe_model_repr(obj):
     if hasattr(obj, '_repr_model_') and six.callable(obj._repr_model_):
         return obj._repr_model_()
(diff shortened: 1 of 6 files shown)


girder_worker/utils.py
```diff
@@ -10,6 +10,10 @@
 import six
 
 
+def is_internal(task):
+    return task.startswith('celery')
```
cjh1: This seems a little fragile? Shouldn't we at least search for the '.' as well? Why not search in set of names, as in ```task in BUILTIN_CELERY_TASKS```?

danlamanna: Can we just check `task.__module__` for `celery.builtins` or something?

girder_worker/utils.py
```diff
@@ -10,6 +10,22 @@
...
+    'celery.chain',
+    'celery.chord',
+    'celery.chord_unlock',
+    'celery.chunks',
+    'celery.group',
+    'celery.map',
+    'celery.starmap']
+
+
+def is_internal(task):
```
danlamanna: `is_celery_task` or `is_builtin_task`?

PR #289, opened 2018-04-12
GirderFolderIdToVolume transform

zachmullen:


kotfic:
@zachmullen this seems generically useful,  do you think we could open this against [girder/girder_worker_utils](https://github.com/girder/girder_worker_utils) in the girder_io file? perhaps with the addition of some tests (e.g. like [these](https://github.com/girder/girder_worker_utils/blob/master/girder_worker_utils/tests/girder_io_test.py))?

kotfic:
Gah! i'm sorry,  you're absolutely right this should be here (along with the other volume stuff).  A test in `test_docker_transforms.py` would be great.  As well as a quick note in the CHANGELOG,  otherwise this looks good to me

zachmullen:
@kotfic test added, PTAL

PR #292, opened 2018-05-17
Allow uploading from a docker volume to a girder folder

zachmullen:
Depends on girder/girder_worker_utils#23


zachmullen:
The upstream change has been merged, so this can now be merged as well. PTAL

PR #293, opened 2018-05-23
Use six to add meta class

cjh1:


kotfic:
@cjh1 Can you try adding the following to `tests/integration/scripts/ansible.cfg`

```
remote_tmp = $HOME/.ansible/tmp
local_tmp = $HOME/.ansible/tmp
```

under the `[default]` section?

kotfic:
@cjh1 Looks like travis is still failing because of some doc-test errors?  do you have any insight as to why? I'm not very familiar with sphinx-doctests

girder_worker/__init__.py
```diff
@@ -1,7 +1,8 @@
 import abc
 import os
 from pkg_resources import DistributionNotFound, get_distribution
-from six.moves.configparser import SafeConfigParser, add_metaclass
+from six.moves.configparser import SafeConfigParser
+from six import add_metaclass
```
cjh1: Oops that would do it!

cjh1:
@kotfic Sorry I'm not familiar with sphinx-doctests, but will take a look.

cjh1:
@kotfic Can you approve this before Lenna moves again ? :smile_cat:

PR #297, opened 2018-07-13
[WIP] Use multistage builds to install sdist in docker container

kotfic:
This uses multistage builds to copy the entire docker context into an
intermediate 'build' container. There we generate an sdist which is
copied into the final container where it is installed. This prevents
unfortunate artifacts like .git/ repos etc laying around and gets us
away from editable installs.

Currently this will break the integration testing infrastructure,  I will be fixing that next

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial
+FROM ubuntu:xenial as base
 
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


zachmullen:
Just verified that I am able to build and run from `Dockerfile` on 00151c9 but not on master.

kotfic:
@zachmullen @danlamanna PTAL

PR #300, opened 2018-08-16
Add ability to upload job artifacts from docker

zachmullen:
Depends on https://github.com/girder/girder_worker_utils/pull/26

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -54,15 +54,10 @@ def girder_before_task_publish(sender=None, body=None, exchange=None,
     try:
         context = get_context()
         if 'jobInfoSpec' not in headers:
-            job = context.create_task_job(Task.girder_job_defaults(),
-                                    sender=sender, body=body,
-                                    exchange=exchange,
-                                    routing_key=routing_key,
-                                    headers=headers,
-                                    properties=properties,
-                                    declare=declare,
-                                    retry_policy=retry_policy,
-                                    **kwargs)
+            job = context.create_task_job(
+                Task.girder_job_defaults(), sender=sender, body=body, exchange=exchange,
+                routing_key=routing_key, headers=headers, properties=properties, declare=declare,
+                retry_policy=retry_policy, **kwargs)
 
         if 'girder_api_url' not in headers:
             context.attach_girder_api_url(sender=sender, body=body,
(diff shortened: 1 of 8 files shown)


girder_worker/app.py
```diff
@@ -31,8 +31,8 @@
     _update_status,
     is_builtin_celery_task,
     is_revoked
-
 )
+from girder_worker_utils.transform import ResultTransform
```
kotfic: Maybe this is ridiculous, but I'm a little concerned about introducing this dependency on girder_worker_utils here - i'd have to think a little harder to make a cogent argument for why.  Would it be a problem to check if there is a `job` attr on the girder_result_hook object that is set to `None`  rather than directly checking the instance type?

zachmullen: If you think that's a better architecture, I will change it, I leave that decision to you.

Since girder_worker already depends on girder_worker_utils, what's the issue with referencing it here?

kotfic: No you're right,  its probably fine as is

zachmullen:
@kotfic this is ready for review. I took a crack at trying to make a test about adding the job automatically to result hooks, but I wasn't sure the best way to do so.

kotfic:
@zachmullen I think the unit tests are probably sufficient  unless you think there is a high likelyhood that this code will break -  either because there is some brittleness or this features use will be uncommon enough that someone might step on it accidentally.  

If so  you could take a look at writing an [integration test](https://github.com/girder/girder_worker/tree/master/tests/integration).  There are a few examples that work some of the transformations in [tests/integration/integration_test_endpoints/server/docker.py](https://github.com/girder/girder_worker/blob/master/tests/integration/integration_test_endpoints/server/docker.py) and the tests that hit those endpoints at [tests/integration/test_docker.py](https://github.com/girder/girder_worker/blob/master/tests/integration/test_docker.py). Let me know if you'd like to write an integration test,  otherwise i'm happy to approve.

zachmullen:
> I think the unit tests are probably sufficient unless you think there is a high likelyhood that this code will break - either because there is some brittleness or this features use will be uncommon enough that someone might step on it accidentally.

I don't think there is, I just want to avoid regressions and the section of code in question [is not covered at all](https://codecov.io/gh/girder/girder_worker/src/31c28c6db32f56e0a6528cbbc8e38c3000d715e5/girder_worker/app.py#L89). But the point of unit tests is that they are cheap both to run and to write; if there isn't a simplistic way for us to unit test this feature, I'm happy to punt on it so we can get this merged. I can open an issue for doing an integration test with result hooks on a future topic.

PR #305, opened 2018-08-28
Call exception hooks on result transforms

zachmullen:
Also make the GirderUploadVolumePathJobArtifact transform
optionally perform the upload if an exception occurs.


kotfic:
Looking at this now i realize,  we should probably be catching the exception and passing it into the function.   Can you file an issue and assign it me?  I'll coordinate updating this code and girder_worker_utils?

zachmullen:
Yeah, I think you're right. They could use `sys.exc_info()`, but it's really a better design to pass it directly.

PR #307, opened 2018-09-04
Fix nvidia fallback.

manthey:
If a docker container is an nvida container (it can be run as an nvidia job), but nvidia was unavailable, it should fail and then try again as a non-nvidia job.  PR #303 broke this behavior.  This fixes it.

### DIFF
--- a/girder_worker/docker/tasks/__init__.py
+++ b/girder_worker/docker/tasks/__init__.py
@@ -3,7 +3,7 @@ import shutil
 import os
 try:
     import docker
-    from docker.errors import DockerException
+    from docker.errors import DockerException, APIError
     from girder_worker.docker import nvidia
     from requests.exceptions import ReadTimeout
 except ImportError:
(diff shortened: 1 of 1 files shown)


girder_worker/docker/tasks/__init__.py
```diff
@@ -49,13 +49,18 @@ def _run_container(image, container_args,  **kwargs):
...
     if runtime is None and nvidia.is_nvidia_image(client.api, image):
         runtime = 'nvidia'
 
     logger.info('Running container: image: %s args: %s runtime: %s kwargs: %s'
                 % (image, container_args, runtime, kwargs))
     try:
-        return client.containers.run(image, container_args, runtime=runtime, **kwargs)
+        try:
+            return client.containers.run(image, container_args, runtime=runtime, **kwargs)
+        except APIError:
```
zachmullen: I'm assuming this is not the exception type raised if, e.g. the docker container crashes or returns non-zero, right?

manthey: It looks like `APIError` gets subclassed as the `NotFound` error as well, which can get raised in other instances.  The more specific error is `APIError: 400 Client Error: Bad Request ("Unknown runtime specified nvidia")`.  In the event that a different error is received, this will try again (without nvidia), which will likely result in the same failure a second time.

girder_worker/docker/tasks/__init__.py
```diff
@@ -49,13 +49,18 @@ def _run_container(image, container_args,  **kwargs):
...
+            if origRuntime is None and runtime is not None:
+                return client.containers.run(image, container_args, **kwargs)
```
zachmullen:
```
else:
    raise
````

girder_worker/docker/tasks/__init__.py
zachmullen: To follow up on this, the things I asked about raise the more generic `DockerException` type, of which `APIError` is a subclass.

PR #309, opened 2018-09-06
Remove doc todos

zachmullen:
This removes a few extraneous todos and also fills in some basics for the "Using Girder Worker with Girder". That text is by no means intended to be comprehensive. I'll open some issues to flesh it out more in the future, but at least now it isn't totally blank.


zachmullen:
@kotfic do you have time to review this?

PR #311, opened 2018-09-17
Move the remote worker plugin from Girder

jbeezley:
I still need to move the tests over, but this appears to be working as is.  One thing I'm not sure about is where to put this inside the package.  I put it inside a local `girder` directory, but that might be confusing and/or require `from __future__ import absolute_import` to work correctly on 2.7.  What would be a good namespace for this? `girder_plugin`?


zachmullen:
We should preserve the history of the girder plugin using `git filter-branch` or @jcfr 's `git-rocket-filter` tools and push the filtered history to this branch.

jbeezley:
I imported the history from girder.  It's a little (or maybe a lot) ugly because files were moved around so much. I did the following:
* filtered the commits to just those changes that touched `plugins/worker`
* grafted `plugins` in `girder` to `girder-plugin` in `girder_worker` to offer a better signal of where the commits came from
* rebased my branch here and deleted `girder-plugin` in the first commit

zachmullen:
This is good to merge, I just want to make sure: do we have an automated integration test that installs the plugin from here into a Girder environment and runs something through the worker?

jbeezley:
https://github.com/girder/girder_worker/pull/311/files#diff-203c8a161b26df5dc98248da978a2c04

but no client tests.

jbeezley:
I guess that mocks out the celery part.  We could do a more end-to-end thing in the integration tests directory, but I'll leave that for future work.

PR #312, opened 2018-09-21
Expose timeout parameter

dorukozturk:
Set the timeout kwarg on the docker python client based on an environment variable (DOCKER_CLIENT_TIMEOUT). Must be a number otherwise it would be ignored.

### DIFF
--- a/girder_worker/docker/tasks/__init__.py
+++ b/girder_worker/docker/tasks/__init__.py
@@ -46,7 +46,11 @@ def _pull_image(image):
 
 def _run_container(image, container_args,  **kwargs):
     # TODO we could allow configuration of non default socket
-    client = docker.from_env(version='auto')
+    if 'DOCKER_CLIENT_TIMEOUT' in os.environ:
+        timeout = int(os.environ['DOCKER_CLIENT_TIMEOUT'])
+        client = docker.from_env(version='auto', timeout=timeout)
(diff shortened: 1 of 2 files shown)


girder_worker/docker/tasks/__init__.py
```diff
@@ -46,7 +46,11 @@ def _pull_image(image):
 
 def _run_container(image, container_args,  **kwargs):
     # TODO we could allow configuration of non default socket
-    client = docker.from_env(version='auto')
+    try:
+        timeout = int(os.environ.get('DOCKER_CLIENT_TIMEOUT'))
+        client = docker.from_env(version='auto', timeout=timeout)
+    except (ValueError, TypeError):
+        client = docker.from_env(version='auto')
```
zachmullen: Suggest:

```
if 'DOCKER_CLIENT_TIMEOUT' in os.environ:
    timeout = int(os.environ['DOCKER_CLIENT_TIMEOUT'])
    client = docker.from_env(version='auto', timeout=timeout)
else:
    client = docker.from_env(version='auto')
```

That way if a user accidentally has a typo in the env var, it'll be an error instead of silently ignoring the env var.

dorukozturk: Sure that sounds better.

PR #314, opened 2018-10-01
[WIP] Add automatic CLI generation from GW tasks

kotfic:
This PR relies on https://github.com/girder/girder_worker_utils/pull/30 and provides a proof of concept implementation that relies on a terrible argument metadata API for auto-generation of a girder worker CLI.  Consider the following girder worker task:

```python
import click
from girder_worker.app import app
from girder_worker_utils.decorators import parameter


@parameter('a', cli_opts={'type': click.INT})
@parameter('b', cli_opts={'type': click.INT})
@parameter('zero',
           cli_args=('-z', '--zero'),
           cli_opts={'type': click.BOOL,
                     'help': 'return "undefined" if dividing by zero'})
@app.task(bind=True)
def divide(self, a, b, zero=False):
    if zero and b == 0:
        return 'undefined'

    return a / b
```

This implements a gwrun command that automatically discovers installed tasks and provides a click based application. E.g.: 

```sh
⇒ gwrun --help
Usage: gwrun [OPTIONS] COMMAND [ARGS]...

Options:
  -o, --output [json|stdout]
  --help                      Show this message and exit.

Commands:
  divide    Proxy that evaluates object once.

```
And

```sh
 ⇒ gwrun divide --help
Usage: gwrun divide [OPTIONS] A B

  Proxy that evaluates object once.

  :class:`Proxy` will evaluate the object each time, while the promise will
  only evaluate it once.

Options:
  -z, --zero BOOLEAN  return "undefined" if dividing by zero
  --help              Show this message and exit.

```

Finally

```sh
⇒ gwrun divide 2 2     
1.0
```

The `cli_args` and `cli_opts` implementation (which is obviously ugly and confusing) are designed to be the low-level API on which a more intuitive and "intelligent" API may be built.  For instance:

```python
@parameter('a', type=float)
@parameter('b', type=float)
@app.task
def add(a, b):
    return a + b
``` 
represents a much nicer API and a mapping could be used to negotiate the difference between `type=float` and `cli_opts={'type': click.FLOAT}` behind the scenes. This serves the purpose of:

1. Improving the readability of the API for the majority usecase
2. Supporting a similar abstraction for handling automatic generation of a web UI 

_Note_ that this also attempts to implement basic output handling whereby the return value of the function may be directly printed to stdout (e.g. \_\_repr\_\_) or serialized to json before being printed.

### DIFF
--- a/girder_worker/cli/__main__.py
+++ b/girder_worker/cli/__main__.py
@@ -1,0 +1,94 @@
+import click
+import pkg_resources as pr
+from click.core import Command, Argument, Option
+from girder_worker.entrypoint import get_extensions, get_extension_tasks, import_all_includes
+from stevedore import driver
+from girder_worker_utils.decorators import (
+    GWFuncDesc,
+    Arg,
+    KWArg)
+
+GWRUN_ENTRYPOINT_GROUP = 'gwrun_output_handlers'
+
+def _cast_to_command(f):
+    if isinstance(f, Command):
+        return f
+
+    description = GWFuncDesc.get_description(f)
+    if description is not None:
+        for arg in reversed(description.arguments):
+            if isinstance(arg, KWArg):
+                if not hasattr(arg, 'cli_args'):
+                    cli_args = ('-{}'.format(arg.name), )
+                else:
+                    cli_args = arg.cli_args
+
+                if not hasattr(arg, 'cli_opts'):
+                    cli_opts = {}
+                else:
+                    cli_opts = arg.cli_opts
+
+                click.option(*cli_args, **cli_opts)(f)
+            elif isinstance(arg, Arg):
+                if not hasattr(arg, 'cli_args'):
+                    cli_args = (arg.name, )
+                else:
+                    cli_args = arg.cli_args
+
+                if not hasattr(arg, 'cli_opts'):
+                    cli_opts = {}
+                else:
+                    cli_opts = arg.cli_opts
+
+                click.argument(*cli_args, **cli_opts)(f)
+
+            else:
+                pass
+    return click.decorators.command()(f)
+
+
+def _iterate_tasks():
+    import_all_includes()
+    for extension in get_extensions():
+        if extension not in ('core', 'docker'):
+            for task in get_extension_tasks(extension).values():
+                yield task
+
+
+
+class GWCommand(click.MultiCommand):
+    def list_commands(self, ctx):
+        return [_cast_to_command(task).name for task in _iterate_tasks()]
+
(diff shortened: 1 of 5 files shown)


kotfic:
To test this I recommend the following:

```sh
mkdir gwruntest; cd gwruntest

# Make a virtualenv
mkvirtualenv -a . gwruntest

# Clone the repos
git clone -b gwrun  git@github.com:girder/girder_worker.git
git clone -b gwrun_argspec git@github.com:girder/girder_worker_utils.git
git clone https://github.com/kotfic/gw_math_tasks.git

# Install the repos
pip install -e girder_worker_utils
pip install -e girder_worker
pip install -e gw_math_tasks

# Test the command
gwrun add 2 2
gwrun --help
```

__Note__ there is an implementation of a toy plugin at [kotfic/gw_math_tasks](https://github.com/kotfic/gw_math_tasks/blob/master/gw_math_tasks/tasks.py)

girder_worker/cli/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+    Arg,
+    KWArg)
+
+GWRUN_ENTRYPOINT_GROUP = 'gwrun_output_handlers'
+
+def _cast_to_command(f):
+    if isinstance(f, Command):
+        return f
+
+    description = GWFuncDesc.get_description(f)
```
jbeezley: Is there a way to make a default description that isn't:
```
  Proxy that evaluates object once.

  :class:`Proxy` will evaluate the object each time, while the promise will
  only evaluate it once.
```

girder_worker/cli/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+    if description is not None:
+        for arg in reversed(description.arguments):
```
jbeezley: Why not loop over `description.positional_args` and `description.keyword_args`?

girder_worker/cli/__main__.py
```diff
@@ -0,0 +1,94 @@
...
+                pass
+    return click.decorators.command()(f)
+
+
+def _iterate_tasks():
+    import_all_includes()
+    for extension in get_extensions():
+        if extension not in ('core', 'docker'):
+            for task in get_extension_tasks(extension).values():
+                yield task
```
jbeezley: I get the convenience of just using the task names, but I do think it will be necessary to at least give the user's a signal about what plugin added the task.

jbeezley:
From a high level, this looks pretty nice.  I think I want to try something non-trivial with it before final judgement though.

manthey:
Stale

PR #319, opened 2018-10-10
RFC: Add retry option to GirderFileIdToVolume

zachmullen:
This bakes download retry directly into the `GirderFileIdToVolume` transform. This isn't necessarily designed to be merged into master, this is more of a test branch and a place for discussion of how best to achieve retry logic for our Girder requests from the workers.

### DIFF
--- a/girder_worker/docker/transforms/girder.py
+++ b/girder_worker/docker/transforms/girder.py
@@ -1,6 +1,7 @@
 import errno
 import os
 import shutil
+import time
 from girder_worker_utils.transform import Transform
 from girder_worker_utils.transforms.girder_io import (
     GirderUploadJobArtifact,
@@ -8,6 +9,7 @@ from girder_worker_utils.transforms.girder_io import (
     GirderUploadToFolder,
     GirderUploadToItem
 )
+from requests.exceptions import RequestException
 from . import TemporaryVolume, Connect, NamedOutputPipe, _maybe_transform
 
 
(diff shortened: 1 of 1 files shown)


zachmullen:
Depends on https://github.com/girder/girder/pull/2839

zachmullen:
ping @danlamanna

kotfic:
I've pinged @danlamanna about girder/girder#2839   I'd like to see a test for this behavior,  but as written the backoff will lag the tests.  Would you consider something like `GirderFileIdToVolume.retry_delay = 1`  and changing th sleep to `time.sleep(i * GirderFileIdToVolume.retry_delay)`   Then write a test that sets retry_delay to 0.01 or something and verifies that max_tries=2  calls downloadFile twice?

zachmullen:
That may be a useful feature, but if we are just thinking about testing, wouldn't it be better to just mock `time.sleep`?

kotfic:
oh sure,  that is a better idea =P

kotfic:
just out of curiosity,  does it make sense to move this functionality into ```downloadFile(...)``` and just pass `max_tries` through the transform?  That way other girder client functions could use it,  and other resource transforms could pick up the benifit?

zachmullen:
I had a discussion about where the retry logic should live with @danlamanna . Putting it here was expedient, but I don't even intend to merge this necessarily, it was designed to generate discussion and opinions on where this should live.

I'm inclined to agree with you that it should really go into girder_client.

PR #321, opened 2018-10-19
Expose some config behavior to utils.

manthey:
Move the logic for getting the temporary directory root to the utils module.  Move the logic for determine if direct paths are allowed to the utils module.

This will allow girder_worker_utils to access these values without using the girder_worker.config object.


manthey:
We can use the TMPDIR environment variable to direct to a specific temporary directory.  For new style tasks, it has been decided that we should also use an environment variable to indicate that direct paths are permitted.  As such, this PR is no longer useful.

PR #322, opened 2018-10-19
Add a guard around getting a local file path.

manthey:
This is necessary for old-style girder_worker jobs using non-filesystem assetstores.

### DIFF
--- a/girder_worker/girder_plugin/utils.py
+++ b/girder_worker/girder_plugin/utils.py
@@ -96,11 +96,9 @@ def girderInputSpec(resource, resourceType='file', name=None, token=None,
         # that location.  This can permit the user of the specification to
         # access the file directly instead of downloading the file.
         try:
-            direct_path = File().getLocalFilePath(resource)
+            result['direct_path'] = File().getLocalFilePath(resource)
         except FilePathException:
-            direct_path = None
-        if direct_path is not None:
-            result['direct_path'] = direct_path
+            pass
     return result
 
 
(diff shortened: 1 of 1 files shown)


girder_worker/girder_plugin/utils.py
```diff
@@ -94,7 +95,10 @@ def girderInputSpec(resource, resourceType='file', name=None, token=None,
         # If we are adding a file and it exists on the local filesystem include
         # that location.  This can permit the user of the specification to
         # access the file directly instead of downloading the file.
-        direct_path = File().getLocalFilePath(resource)
+        try:
+            direct_path = File().getLocalFilePath(resource)
+        except FilePathException:
+            direct_path = None
         if direct_path is not None:
             result['direct_path'] = direct_path
```
zachmullen:
```python
try:
    result['direct_path'] = File().getLocalFilePath(resource)
except FilePathException:
    pass
```

manthey: `direct_path` isn't defined if the exception is thrown.  This could be:
```
        try:
            direct_path = File().getLocalFilePath(resource)
            if direct_path is not None:
                result['direct_path'] = direct_path
        except FilePathException:
            pass
```
is you think that is better.

zachmullen: Ah, I didn't realize `getLocalFilePath` could return `None`. In that case, the way it's written is fine.

What is the semantic difference between returning `None` and raising `FilePathException`?

manthey: In looking at Girder code, it looks like it never returns `None` , so your code is better.  I'll change this and https://github.com/girder/girder/pull/2848.

PR #323, opened 2018-10-24
GirderItemIdToVolume transform

zachmullen:


girder_worker/docker/transforms/girder.py
```diff
@@ -168,6 +168,57 @@ def _repr_model_(self):
...
+            os.makedirs(path)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
+
+        return self._item_id, path
+
+    def transform(self, **kwargs):
+        self._volume.transform(**kwargs)
+        dir = self._volume.host_path
```
kotfic: can we name this something other than built-in?

kotfic:
Other than the `dir` thing this all looks good to me,  I'd like to see an integration test with an item and two files that makes sure both files are downloaded. You might want to make a `test_item_with_files` fixture

zachmullen:
@kotfic my integration test is now succeeding, but a different one is failing, and I have no idea why, any thoughts?

kotfic:
First guess is it has something to do with the  `test_multi_file_item` fixture being set to function scope,  but not cleaning up the item after the `yield`

zachmullen:
All set. The extraneous changes on here are due to chance breakages in two upstream libraries where we always install the latest version: click and flake8.

PR #326, opened 2018-11-01
[WIP] chained docker tasks integration test

zachmullen:


tests/integration/fixtures/docker_test_image/test.py
```diff
@@ -90,5 +90,20 @@ def listdir(d):
...
+@click.argument('out', type=click.File('w'))
+def concat(f1, f2, out):
+    out.write(f1.read() + f2.read())
+
+
+@cli.command()
+@click.argument('f', type=click.File('r'))
+@click.argument('out', type=click.File('w'))
+def reverse(f, out):
+    out.write(reversed(f.read()))
```
42B: Does this line work because of some difference with `click.File('w')`?

```python
In[2]: with open('hmm.txt', 'w') as f:
  ...:     f.write('123\n456\n789\n')
  ...: 
In[3]: with open('hmm.txt', 'r') as f:
  ...:     content = f.read()
  ...: 
In[4]: with open('out.txt', 'w') as f:
  ...:     f.write(reversed(content))
  ...: 
  File "", line 3265, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-844f453fde68>", line 2, in <module>
    f.write(reversed(content))
TypeError: write() argument must be str, not reversed
```

Maybe `out.write(f.read()[::-1])` ?

zachmullen: You're right, I hadn't tested this code yet.

```bash
$ echo "hello" > test.txt
$ docker run -v `pwd`:/data girder/girder_worker_test:ng reverse /data/test.txt /data/out.txt
  File "./test.py", line 109, in <module>
    cli(obj={})
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "./test.py", line 105, in reverse
    out.write(reversed(f.read()))
TypeError: expected a string or other character buffer object
```

You got to this PR so fast I thought you were a robot. In fact I'm still not sure you aren't 😄

42B: @zachmullen 

Haha that's the second time today I've been thought of as a bot! No, I just like looking at all the new python pull requests and am simply trying to learn by exposing myself to a wide variety of projects!

* https://github.com/pulls?page=2&q=is%3Aopen+is%3Apr+language%3Apython&utf8=%E2%9C%93

zachmullen: Haha, well thank you for looking out and catching my mistake!

42B: No problem! At first I thought it was a new way or something so I just tested out a quick example but I didn't have `click` installed in my current PyCharm project so I wasn't 100% sure.

zachmullen: FWIW, to get it working with the `reversed` builtin it would have been:

    ''.join(reversed(f.read()))

girder_worker/docker/tasks/__init__.py
```diff
@@ -376,3 +377,17 @@ def docker_run(task, image, pull_image=True, entrypoint=None, container_args=Non
     return _docker_run(
         task, image, pull_image, entrypoint, container_args, volumes,
         remove_container, **kwargs)
+
+
+@app.task(base=DockerTask, bind=True)
+def docker_run_chained(task, results, container_args=None, *args, **kwargs):
+    container_args = container_args or []
+    kwargs['container_args'] = [_maybe_transform_chain_result(a, results) for a in container_args]
```
kotfic: I'm a little confused.  You shouldn't need to transform these results in the actual task.  The transformation happens when the task class (`girder_worker.task.Task`) calls `__call__` [here](https://github.com/girder/girder_worker/blob/master/girder_worker/task.py#L149-L170)

Ideally we'd have a ResultTransform associated with the first docker task that returns something like a `GirderFileId` transform.  That should get transformed as apart of the above code block before being passed to the second docker_run task. Is there something i'm missing that is preventing that approach?

girder_worker/docker/transforms/girder.py
```diff
@@ -298,3 +298,9 @@ def transform(self, *args, **kwargs):
     def exception(self):
         if self._upload_on_exception:
             return self.transform()
+
+
+class ChainedResultItem(ChainedResultTransform):
+    def transform(self, results):
+        item_id = super(ChainedResultItem, self).transform(results)
```
kotfic: oh i see... do you have multiple results being returned from the first docker task? What are they?

zachmullen: I'm not sure how this actually works, but my understanding of girder_result_hooks was that the return value from a docker_run task was always going to be a list, and I thought it was a list whose values are the resolved result hooks. So in the right-hand task in a chain, we need to be able to select which index into that list is the result we want to use.

I have no idea whether this will actually work. If you have an architecture you think is better, that would be great, but you'd need to write it out in code for me to understand it probably. The code inside `test_docker_run_multistep_workflow` in this PR is where we can show how this looks from an API perspective.

manthey:
Stale

Issue #327, opened 2018-11-30
spaces in girder folder name are incorrectly converted to container args.


subdavis:
Folder name: `UTM 2018-11-21T14:53:51.104Z`

_docker_run:
```
(Pdb) pp container_args
['/mnt/girder_worker/7723d0f170c1436b91deae01a83511d4/5bf5717f89f99f119ce51703/UTM '
 '2018-11-21T14:53:51.104Z',
 '/mnt/girder_worker/7723d0f170c1436b91deae01a83511d4/5bfea50989f99f119ce51d80/variables.csv',
 '--workingfolder',
 '/mnt/girder_worker/7723d0f170c1436b91deae01a83511d4/__results__']
```

Error seen in celery: `cannot open file '2018-11-21T14:53:51.104Z': No such file or directory`

subdavis:
@kotfic 

```
(Pdb) pp image
'samuelgerber/utm'
(Pdb) pp entrypoint
None
(Pdb) pp volumes
{'/tmp/tmpqbclifh4': {'bind': '/mnt/girder_worker/faa3b3626a104b19a115b7579abae1f6',
                      'mode': 'rw'}}
(Pdb) pp container_args
['/mnt/girder_worker/faa3b3626a104b19a115b7579abae1f6/5bf5717f89f99f119ce51703/UTM '
 '2018-11-21T14:53:51.104Z',
 '/mnt/girder_worker/faa3b3626a104b19a115b7579abae1f6/5bfea50989f99f119ce51d80/variables.csv',
 '--workingfolder',
 '/mnt/girder_worker/faa3b3626a104b19a115b7579abae1f6/__results__']
```

kotfic:
@zachmullen i'm working on a failing test now

kotfic:
Turns out the string `'/mnt/girder_worker/faa3b3626a104b19a115b7579abae1f6/5bf5717f89f99f119ce51703/UTM 2018-11-21T14:53:51.104Z'` is not actually being split on the whitespace,  **pdb** is choosing to split the string when ***printing***.  The issue is the underlying container entrypoint script does not handle paths with spaces.

zachmullen:
For posterity, the entrypoint in question is [this R script](https://github.com/KitwareMedical/UTM/blob/master/Scripts/run.utm.barycenter.R).

I don't know anything about R arg parsing, but from a quick glance, I'm guessing it's actually the result of how that R script makes subprocess calls internally to other R scripts (via the `system2` command).

PR #328, opened 2018-12-03
Add a test with appropriate mocking for docker run

kotfic:
I put this test together while investigating https://github.com/girder/girder_worker/issues/327 Many more tests should probably be written,  but this at least demonstrates how to fully mock out the `docker_run` task for posterity's sake


manthey:
Stale

PR #342, opened 2019-10-22
Fix/jinja2 security issue

kotfic:


requirements3.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file requirements3.txt requirements.in requirements-dev.in
+#    pip-compile --output-file=requirements3.txt requirements-dev.in requirements.in
```
zachmullen: Does this change in order matter?

kotfic: i don't think so,  in any case its autogenerated by pip-compile so i'd be hesitant to touch it

zachmullen: I assumed the auto-generator just dumped the value of `' '.join(sys.argv)`, and that you had run it in a different order, but it seems to me that the output should be stable, as the input files should be treated as a set.

PR #344, opened 2019-10-23
Run integration tests nightly rather than on each commit

kotfic:


.circleci/config.yml
```diff
@@ -167,6 +167,18 @@ jobs:
 
 workflows:
   version: 2
+  nightly:
+    triggers:
+      - schedule:
+          cron: "17 2 * * *"
```
zachmullen: I'm getting into numerology trying to guess what these numbers are about. They're both prime... that must be important, right?

kotfic: everyone runs at "0 0 * * *" which overloads their servers at 12:00am, they literally have documentation that is like,  please run at some other time.  We're running at 2:17am which was chosen at random

zachmullen: Good stuff. IMO CircleCI should publish an option that lets you just say "run it whenever you have the cycles, within some period of hours". I think most people would probably be fine with that.

PR #349, opened 2020-01-06
Call cleanup on girder_result_hook transforms. Fixes #348

### LINKED ISSUE #348: girder_result_hooks cleanup methods are not called after task is done
Hello everyone,

There seems to be an issue with task cleanup. The cleanup methods for transforms in args and kwargs are being called properly at the end of the task, but that is not the case for transforms in girder_result_hooks. Is this intended ?

This means for instance that when passing GirderUploadToItem in girder_result_hooks with the delete_file flag as True, the file is never deleted.

I'm pushing a potential fix for this issue:
https://github.com/girder/girder_worker/pull/349

boucaud:
https://github.com/girder/girder_worker/issues/348

### DIFF
--- a/girder_worker/task.py
+++ b/girder_worker/task.py
@@ -168,3 +168,6 @@ class Task(celery.Task):
         finally:
             _walk_obj(args, self._maybe_cleanup)
             _walk_obj(kwargs, self._maybe_cleanup)
+            if hasattr(self.request, 'girder_result_hooks'):
+                for hook in self.request.girder_result_hooks:
+                    self._maybe_cleanup(hook)


kotfic:
Yeah this looks good to me,  thanks for the fix!

kotfic:
@boucaud Sorry looks like we neglected to turn on CircleCI for forked PRs,  could you push a new commit to trigger?  Thanks!

zachmullen:
`git commit -m "kick CI" --allow-empty` should work

boucaud:
These CI failures don't seem related, WDYT ?

kotfic:
definitely not your issue,  its an issue with code coverage major version mismatch in the tox file and requirements*.txt file.  I am seeing if I can run down the "right" solution before I pin coverage<5.0.0

kotfic:
Hi @boucaud,  I've updated the testing in master to resolve the coverage issue.  Can you rebase on master?  Thanks for hanging in there!

boucaud:
Seems fine now, thanks !

zachmullen:
Thanks for the fix @boucaud !

PR #352, opened 2020-02-07
"Fix" nightly CI

zachmullen:
@kotfic PTAL, this will only make a difference on the nightly CI


zachmullen:
If we care, we can go do a follow up branch to remove all the infrastructure related to py2 integration testing, but I don't care enough to do so. If anyone asks I'll just say don't use python 2.

PR #355, opened 2020-04-08
correct `girder_client_token` usage example

subdavis:
`girder_client_token` accepts a string, not a document.
`Token()` must be used instead of `Token`


HACKING.md
```diff
@@ -274,11 +274,11 @@ from girder.models.token import Token
 	.param('file_id', 'File Id to run "some_task" on', dataType='string'))
 def some_rest_endpoint(file_id, **params):
 	# Create a token with permissions of the current user.
-	token = Token.createToken(
+	token = Token().createToken(
 		user=self.getCurrentUser())
 
 	# Launch the asynchronous task
-	async_result = some_task.delay(file_id, girder_client_token=token)
+	async_result = some_task.delay(file_id, girder_client_token=str(token['_id']))
```
zachmullen, line 281: The cast to `str` is actually unnecessary here, as the token's `_id` is already a string.

subdavis: Ah, so it is.  I'm accustomed to any `_id` being a bson object id.  Looks like there's some python3 issue to sort out anyway.  I'll take a look tomorrow.

kotfic:
@subdavis if you want to drop the `str` cast this looks good to me. thanks for the update!

subdavis:
@kotfic done, ty.

PR #359, opened 2020-06-16
Remove python 2 from CI

zachmullen:


zachmullen:
@kotfic PTAL, this will unblock other contributions once merged.

PR #361, opened 2020-06-24
Fail to start if exception occurs during plugin load

### LINKED ISSUE #360: Exceptions are silently ignored during plugin load
If an exception happens on the worker during plugin load time, the traceback of the exception is not printed as expected on [this line](https://github.com/girder/girder_worker/blob/df00a6f9ffa6d325bcb03a95793858d87fc22704/girder_worker/entrypoint.py#L55).

The message `Problem loading plugin mouse_pain_worker, skipping` does appear.

zachmullen:
Fixes #360

@kotfic PTAL. I guess celery is doing some dark magic monkey patching of `print` and standard pipes...

### DIFF
--- a/girder_worker/entrypoint.py
+++ b/girder_worker/entrypoint.py
@@ -10,27 +10,9 @@ from stevedore import extension
 #: Defines the namespace used for plugin entrypoints
 NAMESPACE = 'girder_worker_plugins'
 
-# Define an internal registry for extensions that aren't associated
-# with an entrypoint.  This registry is primarily useful for testing
-# where one doesn't necessarily want to install a package into the
-# environment for a one off test task.
-_extensions = {}
-
-
-def _import_module(module_name):
-    """Try to import a module given as a string."""
-    module = None
-    try:
-        module = import_module(module_name)
-    except ImportError:
-        import traceback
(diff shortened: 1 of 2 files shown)


zachmullen:
Looks like a test was depending on the wrong behavior. I will try to fix it up later I guess.

zachmullen:
I'm not convinced we should start up if there's an exception during load. Would anyone actually defend that feature?

kotfic:
I think the intention was to allow multiple plugins to be installed and if one fails to not cripple the whole system.  In reality though I think you're probably right,  plugin failure should probably cause a full stop and loudly complain.

zachmullen:
I've changed the scope of this, and it's ready for review. Namely, we now fail catastrophically if plugins fail to load, which is desirable behavior. Aside from that, I also removed the hacked-up extension registry capability. Now the entry point is the exclusive way to load plugins. This gave us the opportunity to remove a fair amount of code.

kotfic:
Oh,  this represents a somewhat big change (failing loudly i mean) will you cut a new minor version once this is in?

zachmullen:
Yes, I plan to cut a release as soon as these two PRs are merged. Thanks!

PR #362, opened 2020-06-25
Un-abstractify base class

zachmullen:
It now provides sensible default behaviors. Plugin entry points now no longer need to provide an `__init__` unless they actually need it. Additionally, plugins no longer strictly need to provide a task listing method, to support plugins that might not expose custom tasks.

### DIFF
--- a/girder_worker/__init__.py
+++ b/girder_worker/__init__.py
@@ -1,8 +1,6 @@
-import abc
 import os
 from pkg_resources import DistributionNotFound, get_distribution
 from six.moves.configparser import SafeConfigParser
-from six import add_metaclass
 
 from . import log_utils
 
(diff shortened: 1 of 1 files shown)


girder_worker/__init__.py
```diff
@@ -33,20 +31,17 @@
 logger = log_utils.setupLogger(config)
 
 
-@add_metaclass(abc.ABCMeta)
 class GirderWorkerPluginABC(object):
```
subdavis, line 34: Should this still be named `ABC` if it's not a true ABC?  Seems like just a regular base class now.

zachmullen: I don't think it should, however I'm not trying to make a breaking change that would be trouble for all downstream plugins. As written, this should be non-breaking.

kotfic:
I'm fine with this,  though it does feel like a ripe opportunity for some primo bike-shedding =P

zachmullen:
Could you approve @kotfic ?

PR #364, opened 2020-07-15
Only send log value if it has data

zachmullen:
Girder's server-side handler writes data to the job log even if it's just an empty string, which ends up writing e.g. `["", "", "", ""]` to the database, growing linearly with each progress update event. It's debatable whether that server side behavior is desirable, however it's easy enough to handle it here and only send the log value if something has been written to it.


zachmullen:
Tagged and pushed this on `v0.7.2`.

PR #366, opened 2020-07-15
Add `python_requires` to setup.py

jwodder:
This package's classifiers indicate that it only supports Python 3.6 and up.  This PR thus sets the package's `Requires-Python` metadata to `>= 3.6` so that pip will refuse to install it on older Python versions (Classifiers by themselves do nothing).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,6 +102,7 @@ setuptools.setup(
         'install': CustomInstall
     },
     install_requires=install_reqs,
+    python_requires='>=3.6',
     zip_safe=False,
     entry_points={
         'console_scripts': [


setup.py
```diff
@@ -102,6 +102,7 @@ def run(self, *args, **kwargs):
         'install': CustomInstall
     },
     install_requires=install_reqs,
+    python_requires='>= 3.6',
```
zachmullen, line 105: The way this is typically written is `>=3.6`, without the space. Could you change that, then I will merge.

Thanks!

jwodder: Space removed.

zachmullen:
Included in release 0.7.3

PR #372, opened 2021-05-13
Cancel inactive jobs immediately

### LINKED ISSUE #254: Cancellation of docker_run jobs does not seem to work
I canceled a batch of docker_run.delay jobs from the Girder jobs UI. The ones that were running just kept running, the ones that were not yet started entered the "Canceling" state but never got out of it.

screen shot 2018-01-22 at 11 54 11 am

subdavis:
Untested, but this is the general idea.

fixes #254

### DIFF
--- a/girder_worker/girder_plugin/event_handlers.py
+++ b/girder_worker/girder_plugin/event_handlers.py
@@ -16,13 +16,13 @@
 
 from __future__ import absolute_import
 
-from celery.result import AsyncResult
-
 from girder import logger
 from girder.exceptions import ValidationException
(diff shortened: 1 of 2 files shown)


zachmullen (approved):
It would be nice if we could make CI green... maybe we can just remove `flake8-blind-except` from our test deps.

subdavis:
It gets installed whether it's part of dev requirements or not.  

Removed in 7217e6c but it still fails.  Some other package must be installing it.

zachmullen (approved):

subdavis:
Tested, working.

PR #375, opened 2022-01-13
Nvidia device requests

manthey:


manthey:
They do different things: `or {}` is in python so if the left side is `None` we have a dictionary.  `|| []` is in javascript so if the left side is `null` or `undefined` we have an array.  The `| 0` is to cast to an integer.

jeffbaumes:
Wow, with the snippets I did not even realize I was looking at two languages! Thanks, makes sense.

PR #376, opened 2022-01-20
Add requests parameters for communication with Girder

pchoisel:
Add a way to specify arguments that will be set to a requests session used for following requests to Girder.

### DIFF
--- a/docs/using-from-girder.rst
+++ b/docs/using-from-girder.rst
@@ -64,3 +64,32 @@ be passed into the function in place of the transform object. For example:
 
     def process_file(file):
         return my_task.delay(input_file=GirderFileId(file['_id'])).job
+
+
+Customizing communication back to Girder server
+-----------------------------------------------
+
+Girder Worker is bound to communicate with a Girder instance to update the information about running jobs or download files from the assetstore.  
+In some scenarios, you might want to customize the HTTP communication back to the Girder server.
+  
+Girder Worker uses the Python library ``requests`` to communicate over HTTP and creates a `requests.Session() <https://docs.python-requests.org/en/master/user/advanced/#session-objects>`_ instance that can be adjusted by providing a ``girder_client_session_kwargs`` kwarg in the task parameters.
+
+Here are examples of using ``girder_client_session_kwargs``:
+
+.. code-block:: python
+
+    kwargs = {
+        'girder_client_session_kwargs': {'verify': False}
+    }
+    task.apply_async((), kwargs)
+
+This will disable TLS certificate validation when communicating with Girder.
+
+.. code-block:: python
+
+    kwargs = {
+        'girder_client_session_kwargs': {'headers': {'x-test': 'true'}}
+    }
+    task.apply_async((), kwargs)
+
+This will set the headers of any subsequent request to `{'x-test': 'true'}`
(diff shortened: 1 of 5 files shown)


HACKING.md
```diff
@@ -299,6 +299,22 @@ def some_task(self, file_id):
 
 In this example the producer side REST endpoint creates a token with the scope of the current user, then passes the REST argument `file_id` and the token to the task object's `.delay(...)` function. On the consumer side, the task has access to the GirderClient instance through `self.girder_client`. To get access to `self` the task must be decorated with `@app.task(bind=True)` (See [bound tasks](http://docs.celeryproject.org/en/latest/userguide/tasks.html#bound-tasks) for more information).
 
+##### What is Girder is served over an HTTPS connection ?
```
zachmullen, line 302:
```suggestion
##### Customizing communication back to Girder server
```

I'm OK with the code changes you've added, but I think we should document it more generally as a feature, i.e. configurability of the request session. `verify: False` could still be used as the example.

Aside from that, it's more accurate to say "What if Girder is served over an _unauthenticated_ TLS connection?" which is kind of a strange use case. What's the reason you need this, out of curiosity?

zachmullen (changes requested):

HACKING.md
pchoisel: Hi Zach, thanks for the review.

I will document it more tomorrow.
We need this for a project where Girder is not the front-ent webserver (an Apache with load-balancer is set up in front in another VM). However, the security consultants hired by the client require us to cypher the communications between Apache and Girder. To do that, a self-signed certificate is created and we don't want to bother with signing it.

zachmullen: That's an... interesting... threat model. Lots of people use tools like [Consul](https://www.consul.io/) for secure inter-service authentication. I'd suggest checking that out if security is something you care about. It manages the hassles of CAs and TLS for you pretty nicely.

pchoisel: Thanks for the suggestion, I'll be sure to suggest it to the client.

I changed my `HACKING.md` section to a feature documentation. Is that what you had in mind ?

docs/using-from-girder.rst
```diff
@@ -64,3 +64,32 @@ be passed into the function in place of the transform object. For example:
 
     def process_file(file):
         return my_task.delay(input_file=GirderFileId(file['_id'])).job
+
+
+Customizing communication back to Girder server
+-----------------------------------------------
+
+Girder Worker is bound to communicate with a Girder instance to update the information about running jobs or download files from the assetstore.  
+In complex platforms, you might want to customize the HTTP communication back to the Girder server.
```
zachmullen, line 73:
```suggestion
In some scenarios, you might want to customize the HTTP communication back to the Girder server.
```

zachmullen (changes requested):
LGTM after one small tweak

pchoisel:
There are two unit tests failing, but I have no idea why. Would you have a suggestion ?

girder_worker/utils.py
```diff
@@ -207,6 +218,12 @@ def __init__(self, logPrint, url, method=None, headers=None, interval=0.5,
         self._progressCurrent = None
         self._progressMessage = None
 
+        self._session = requests.Session()
```
zachmullen, line 221: Can we make sure to close `self._session` in `cleanup`?

zachmullen:
The unit test failures appear to be a result of this branch; I just re-ran them on master and they succeeded.

pchoisel:
I had to modify the tests mocking requests.request
All the tests are passing now and I added the session.close().

PR #389, opened 2024-01-03
Added code to support Singularity on HiperGator

suhasthegame:


girder_worker/docker/tasks/__init__.py
```diff
@@ -419,34 +461,109 @@ def _docker_run(task, image, pull_image=True, entrypoint=None, container_args=No
...
+                if job_info in [drmaa.JobState.DONE, drmaa.JobState.FAILED]:
+                    break
+
+                # Check if the task has been aborted by the user
+                if task.is_aborted:
+                    s.control(jobid, drmaa.JobControl.TERMINATE)
+                    logger.info(f'Job {jobid} was cancelled by user.')
+                    return JOB_STATUS.CANCELLED
+
+                time.sleep(5)  # Sleep to avoid busy waiting
```
manthey, line 556: I think it'd be better to call `s.wait(5)` and catch `ExitTimeoutException`.

girder_worker/docker/tasks/__init__.py
```diff
@@ -359,6 +363,44 @@ def _cleanup_temp_volumes(self, temp_volumes, default_temp_volume):
...
+        #check if the user has provided the image of the plugin.
+        if not image:
+            image = "sarderlab/histo-cloud:latest"
+            # raise ValueError('Plugin Image required for Singularity')
+        
+        #Commnad to be called for executing Singularity Job
+        bind_paths[temporary_directory] = '/output'
+        super().__call__(*args,**kwargs)
+        
+        qos = kwargs.pop('qos', 'pinaki.sarder')
```
manthey, line 384: I'd recommend having these from environment variables.  And, rather than construct the slurm script ourselves, we could use the `simple_slurm` package to do it (it does the same thing, but has been validated by others and escapes command line parameters appropriately).

girder_worker/docker/tasks/__init__.py
```diff
@@ -419,34 +461,109 @@ def _docker_run(task, image, pull_image=True, entrypoint=None, container_args=No
...
+            return s.connect_ex('/var/run/docker.sock') != 0
+    except socket.error:
+        return True
+    
+def _generate_slurm_script(image, container_args, bind_paths, qos, cpus, gpus, memory, other_slurm_options):
+    # Construct the bind option for the Singularity command
+    bind_option = ','.join([f"{host}:{container}" for host, container in bind_paths.items()])
+
+    # Construct the Singularity command
+    singularity_command = f'singularity exec --bind {bind_option} docker://{image} {" ".join(container_args)}'
```
manthey, line 517: The join is unsafe -- if the container args contain spaces or quotes or shell characters, odd things will happen.

PR #413, opened 2026-07-28
Change how we find the worker's network

manthey:
We enumerate docker containers to find the primary worker's network so we can attach jobs to the same network, allowing them to reference girder the same way as the worker.  If the containers mutate while we are listing them, this can throw can an error an prevent network location.  The docker python script has two methods of getting the container list; one is an iterator and the other is just a list() of the iterator.  But if the iterator throws while iterating, we don't don't anything from the list(), whereas if we use the iterator we can still proceed.

### DIFF
--- a/girder_worker/docker/tasks/__init__.py
+++ b/girder_worker/docker/tasks/__init__.py
@@ -49,6 +49,10 @@ def _pull_image(image):
         raise
 
 
+def _is_no_such_container(exc):
+    return 'No such container' in str(exc) or type(exc).__name__ == 'NotFound'
+
+
 def _get_docker_network():
     try:
         ip = socket.gethostbyname(socket.gethostname())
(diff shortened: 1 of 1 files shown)


manthey:
@marySalvi This is part one of upstreaming the mixed up log issue.  If you pull this into your local worker and disable the monkey patch and then confirm that we still have a fix, we can merge this in.

manthey:
We expect with this change that all jobs will be attached to the girder worker network.  The direct symptom before this change is that when jobs are started or stopped with certain times we see log messages in girder_worker that read "Failed to get docker network".  Any job that uses the network (e.g., uses girder_client), could fail if it was not assigned the same network as girder_worker.  With this change, we should never see "Failed to get docker network" log messages.

As a future improvement, we probably could cache this and never recheck the containers after we obtained the network the first time.

marySalvi (approved):

marySalvi:
This was tested with a script that created multiple jobs in quick succession. Without these changes implemented I saw "Failed to get docker network" logged for approximately 2 out of every 100 jobs created. 
With the changes in this PR implemented the failure never happened while creating 1000 jobs.

PR #414, opened 2026-07-28
Make job_manager and girder_client per request

manthey:
We had set the job_manager and girder_client per task, but if we start two tasks at close to the same time this could result in the two tasks sharing the same job manager and girder client, resulting in one task ending up with the output streams from both tasks.  By making these per request, which are always distinct, this avoids the issue.


manthey:
@marySalvi This should address jobs logging to the wrong job.

marySalvi (approved):
This was tested with a script that creates multiple jobs in quick succession that generate log messages. Without these changes many of the jobs had conflated log messages. With these changes all of the log messages were distinct. Tested with 100 jobs created before and after changes each.

PR #416, opened 2026-07-28
When using a celery threads-pool for the workers, do not tee logPrint

manthey:
When we are using a threads pool, tee-ing the logPrint output can lead to the logs being written to multiple threads and therefore showing up to multiple workers

### DIFF
--- a/girder_worker/app.py
+++ b/girder_worker/app.py
@@ -12,7 +12,6 @@ from celery.signals import (
     task_prerun,
     task_revoked,
     task_success,
-    worker_init,
     worker_ready)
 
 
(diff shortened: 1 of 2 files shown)


marySalvi (approved):
This was tested using a script that creates multiple jobs in quick succession. 
**As expected:** 

- The jobs that were using a threads pool printed the message 'Using a threads pool' to the worker container logs at startup but did **not** print header logs to the job log output in the girder gui admin console. 
- The jobs that were not using a threads pool printed the message 'Not using a threads pool' to the worker container logs at startup **and also** printed header logs to the job log output in the girder gui admin console.

